PR #142, opened 2019-10-09
Swagger documentation gen for signing API

marklnichols:


tomsmalley:
Closing because this appears to be superseded by https://github.com/kadena-io/signing-api

PR #143, opened 2019-10-10
Enable preview step for Signing API process

mankyKitty:
This extends the deployment workflow to be able to handle the preview step for signing requests and submit the sign on confirm of the preview stage, regardless of the status of the response from the network.

It is still marked as WIP because there are test buttons and data in the UI code to enable investigation of this workflow. They need to be removed before we merge.

### DIFF
--- a/frontend/src/Frontend/UI/Dialogs/DeployConfirmation.hs
+++ b/frontend/src/Frontend/UI/Dialogs/DeployConfirmation.hs
@@ -72,8 +72,8 @@ data DeployConfirmationConfig t = DeployConfirmationConfig
   { _deployConfirmationConfig_modalTitle :: Text
   , _deployConfirmationConfig_previewTitle :: Text
   , _deployConfirmationConfig_previewConfirmButtonLabel :: Text
-    -- The confirmation button the preview screen will be disabled if the network returns
-    -- an error. Some processes like the signing API need to override this as the
+    -- The confirmation button in the preview screen will be disabled if the network
+    -- returns an error. Some processes like the signing API need to override this as the
     -- responsibility for the transaction being signed lies with the creator, not the
     -- person doing the signing. This function will be called twice, once with the bool
     -- value of the status of the network call, and again with the setting for the
@@ -145,7 +145,7 @@ fullDeployFlow deployCfg model runner onClose =
   fullDeployFlowWithSubmit deployCfg model showSubmitModal runner onClose
   where
     showSubmitModal chain result done _next nodes =
-      pure $ attachWith (\n _ -> Right $ deploySubmit chain result n) nodes done
+      pure $ Right . deploySubmit chain result <$> nodes <@ done
 
 -- | Workflow taking the user through Config -> Preview -> Status
 fullDeployFlowWithSubmit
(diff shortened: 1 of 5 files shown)


frontend/src/Frontend/UI/Dialogs/DeployConfirmation.hs
```diff
@@ -178,24 +229,43 @@ fullDeployFlow cfgTitle model runner _onClose = do
...
+        next <- uiButtonDyn
+          (def & uiButtonCfg_class .~ "button_type_confirm" & uiButtonCfg_disabled .~ isDisabled)
+          $ text (_deployConfirmationConfig_previewConfirmButtonLabel dcfg)
+
         pure (back, next)
 
-      let done = gate (current succeeded) next
+      let done = gate (current $ fmap ignoreSuccessStatus succeeded) next
+
+      (eDoneAfterConfirm, eNextWorkflow) <- either (,never) (never,)
```
tomsmalley: `fanEither`

frontend/src/Frontend/UI/Dialogs/DeployConfirmation.hs
```diff
@@ -64,6 +68,29 @@ import qualified Pact.Types.Command as Pact
 import qualified Servant.Client.JSaddle as S
 import qualified Text.URI as URI
 
+data DeployConfirmationConfig t = DeployConfirmationConfig
+  { _deployConfirmationConfig_modalTitle :: Text
+  , _deployConfirmationConfig_previewTitle :: Text
+  , _deployConfirmationConfig_previewConfirmButtonLabel :: Text
+  , _deployConfirmationConfig_disregardSubmitResponse :: Bool -> Bool
```
tomsmalley, line 86: Can you add a comment explaining this field?

frontend/src/Frontend/UI/Dialogs/DeployConfirmation.hs
mankyKitty: Much nicer, thanks !

frontend/src/Frontend/UI/Dialogs/DeployConfirmation.hs
```diff
@@ -64,6 +68,40 @@ import qualified Pact.Types.Command as Pact
 import qualified Servant.Client.JSaddle as S
 import qualified Text.URI as URI
 
+data DeployConfirmationConfig t = DeployConfirmationConfig
+  { _deployConfirmationConfig_modalTitle :: Text
+  , _deployConfirmationConfig_previewTitle :: Text
+  , _deployConfirmationConfig_previewConfirmButtonLabel :: Text
+    -- The confirmation button the preview screen will be disabled if the network returns
```
alexfmpe: "in" the preview screen?

frontend/src/Frontend/UI/Dialogs/DeployConfirmation.hs
```diff
@@ -99,11 +137,29 @@ fullDeployFlow
...
-  => Text -- ^ Title of modal
+  => DeployConfirmationConfig t
+  -> model
+  -> m (modelCfg, Event t DeploymentSettingsResult)
+  -> Event t () -> m (modelCfg, Event t ())
+fullDeployFlow deployCfg model runner onClose =
+  fullDeployFlowWithSubmit deployCfg model showSubmitModal runner onClose
+  where
+    showSubmitModal chain result done _next nodes =
+      pure $ attachWith (\n _ -> Right $ deploySubmit chain result n) nodes done
```
alexfmpe: Alternatively `pure $ Right . deploySubmit chain result <$> nodes <@ done`

PR #146, opened 2019-10-15
Reorganise wallet setup screens

mankyKitty:
As part of the redesign process, the workflow for the wallet creation / recovery has been updated to have the passwords fields on the same screen as the recovery phrase creation/input.

The workflow has been updated to suit and a small timer has been added on the password validation steps that is gated on both fields be non-empty on a 0.4 second delay. This prevents the validation from spamming the user and also means they don't have to wait until they've clicked next before they get some feedback about the password input.

Removed the `h1` from the password fields but there are no other styling changes so it looks pretty janky at the moment.


desktop/src/Desktop/Setup.hs
```diff
@@ -176,36 +192,46 @@ confirmPhrase mnemonic = Workflow $ do
...
   skip <- uiButton btnCfgTertiary $ text "Skip"
   next <- confirmButton (def & uiButtonCfg_disabled .~ fmap not done) "Next"
-  finishSetupWF $ leftmost
-    [ createNewWallet (Just mnemonic) <$ back
-    , setPassword (createNewWallet Nothing) (Crypto.sentenceToSeed mnemonic Crypto.english "")
-      <$ (gate (current done) next <> skip)
-    ]
+
+  pure
+    ( fmapMaybe id $ current dPassword <@ gate (current done) next <> skip
```
alexfmpe: `tagMaybe (current dPassword) $ gate (current done) next <> skip`

desktop/src/Desktop/Setup.hs
```diff
@@ -176,36 +192,46 @@ confirmPhrase mnemonic = Workflow $ do
...
-    back <- cancelButton (def & uiButtonCfg_type ?~ "button") "Back"
-    next <- confirmButton (def & uiButtonCfg_type ?~ "submit") "Next"
-    let (err, pass) = fanEither $ attachWith (\p _ -> uncurry checkPassword p) (liftA2 (,) p1 p2) next
-    lastError <- holdDyn Nothing $ Just <$> err
-  pure
-    ( Crypto.generate seed . T.encodeUtf8 <$> pass
-    , previous <$ back
-    )
+
+    let (err, pass) = fanEither $ attachWith (\p _ -> uncurry checkPassword p) (liftA2 (,) p1 p2) eCheckPassword
```
alexfmpe: `let (err, pass) = fanEither $ checkPassword p <$> p1 <*> p2 <@ eCheckPassword`

desktop/src/Desktop/Setup.hs
```diff
@@ -133,21 +142,28 @@ createNewWallet mMnemonic = Workflow $ do
       elClass "i" "fa fa-lg fa-refresh" blank
       text " Regenerate"
 
+  ePassword <- setPassword (fmap (fmap (\mn -> Crypto.sentenceToSeed mn Crypto.english "") . hush) mnemonic)
+    >>= holdDyn Nothing . fmap pure
+
+  let nextGate = liftA2 (||) (fmap isNothing dPassword) (fmap not stored)
```
alexfmpe: `gate beh ev` allows `ev` through when `beh` is `True`, so I thought `nextGate` would be `True` when allowing the "next" action. Maybe call it `disabledNext`

tomsmalley (approved):
This looks solid. There are a couple of things that need altering, but I'll create new tickets for that.

Merging into a new `redesign` branch.

backend/sass/common.blocks/_wallet.scss
```diff
@@ -4,6 +4,166 @@
...
+$fullscreen-wallet-grey: #f1f1f1;
+$wallet-top-btm-margin: 0.5rem 0 0.5rem 0;
+$kadena-orange: #F3AB3C;
+
+.fullscreen {
+  display: grid;
+  grid-template-columns: 1fr 1fr 1fr 1fr;
+  grid-template-rows: 0.2fr 1.6fr;
+  grid-template-areas: "back header header ." ". wrapper wrapper .";
+}
```
eskimor, line 16: Sorry, just stumbled upon this. This should really not be here. Either that fullscreen is an element of the wallet block, then it should be named wallet__fullscreen or it is a block of its own. In that case it should have its own file _fullscreen.

tomsmalley: Ah, I totally missed the file name. The setup screen code should probably have a `_setup` file, since it's the only place we have the fullscreen stuff anyway. I'll add a ticket to move it

eskimor: Sounds good! Thanks! :-)

PR #149, opened 2019-10-18
Support signing with capabilities

tomsmalley:
Adds support for capabilities. The signing API passes all capabilities
it needs, and these are displayed to the user with an account dropdown
box. Other deployment methods allow the user to specify arbitrary
capabilities.

The sender/gas payer field has not been altered here, but in future
could be moved to the capabilities/roles section too.

This also fixes the transaction hash stuff in deployment dialogs. We
were hashing the code, not the command.

### **Don't merge this yet, this isn't a guaranteed change upstream**

### DIFF
--- a/frontend/src/Frontend/UI/DeploymentSettings.hs
+++ b/frontend/src/Frontend/UI/DeploymentSettings.hs
@@ -42,6 +42,7 @@ module Frontend.UI.DeploymentSettings
   , Identity (runIdentity)
   ) where
 
+import Debug.Trace (traceShowId)
 import Control.Applicative (liftA2)
 import Control.Arrow (first, (&&&))
 import Control.Error.Util (hush)
@@ -585,15 +586,27 @@ uiChainSelection info cls = mdo
     d <- _dropdown_value <$> dropdown Nothing (mkOptions <$> chains) cfg
     pure d
 
+data CapabilityInputRow t = CapabilityInputRow
+  { _capabilityInputRow_empty :: Dynamic t Bool
+  , _capabilityInputRow_value :: Dynamic t (Maybe (Map AccountName [SigCapability]))
+  , _capabilityInputRow_error :: Event t Bool
+  }
+
+data CapabilityInputRows t = CapabilityInputRows
+  { _capabilityInputRows_anyEmpty :: Dynamic t Bool
+  , _capabilityInputRows_anyError :: Event t Bool
+  , _capabilityInputRows_value :: Dynamic t (Maybe (Map AccountName [SigCapability]))
+  }
+
 -- | Display a single row for the user to enter a custom capability and
 -- account to attach
 capabilityInputRow
   :: MonadWidget t m
   => Maybe DappCap
   -> m (Dynamic t (Maybe AccountName))
-  -> m (Dynamic t Bool, Dynamic t (Maybe (Map AccountName [SigCapability])))
+  -> m (CapabilityInputRow t)
 capabilityInputRow mCap mkSender = elClass "tr" "table__row" $ do
-  (empty, parsed) <- el "td" $ mdo
+  (empty, parsed, errors) <- el "td" $ mdo
     cap <- uiInputElement $ def
       & inputElementConfig_initialValue .~ foldMap (renderCompactText . _dappCap_cap) mCap
       & initialAttributes .~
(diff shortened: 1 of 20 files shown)


frontend/src/Frontend/UI/DeploymentSettings.hs
```diff
@@ -572,21 +579,87 @@ uiChainSelection info cls = mdo
...
+  account <- el "td" mkSender
+  pure
+    ( empty
+    , do
+      empty >>= \case
+        True -> pure $ Just mempty
+        False -> do
+          ma <- account
+          e <- parsed
+          pure $ ffor ma $ \a -> Map.singleton a (either (const []) pure e)
```
alexfmpe: I think `pure $ ffor ma $ \a -> Map.singleton a (toList e)` does the same - unsure of which is clearer.

frontend/src/Frontend/UI/DeploymentSettings.hs
```diff
@@ -572,21 +579,87 @@ uiChainSelection info cls = mdo
...
+  => m (Dynamic t (Maybe AccountName))
+  -> m (Dynamic t Bool, Dynamic t (Maybe (Map AccountName [SigCapability])))
+capabilityInputRow mkSender = elClass "tr" "table__row" $ do
+  (empty, parsed) <- el "td" $ mdo
+    cap <- uiInputElement $ def
+      & initialAttributes .~ "placeholder" =: "(module.capability arg1 arg2)" <> "class" =: "input_width_full"
+      & modifyAttributes .~ ffor errors (\e -> "style" =: ("background-color: #fdd" <$ guard e))
+    empty <- holdUniqDyn $ T.null <$> value cap
+    let parsed = parseSigCapability <$> value cap
+        hasError = (\p e -> isLeft p && not e) <$> parsed <*> empty
```
alexfmpe: Does `empty` imply `parseSigCapability` return `Left` ?

frontend/src/Frontend/UI/DeploymentSettings.hs
```diff
@@ -572,21 +579,87 @@ uiChainSelection info cls = mdo
...
+        errors = leftmost
+          [ tag (current hasError) (domEvent Blur cap)
+          , False <$ _inputElement_input cap
+          ]
+    pure (empty, parsed)
+  account <- el "td" mkSender
+  pure
+    ( empty
+    , do
+      empty >>= \case
```
alexfmpe:
```haskell
ffor3 empty account parsed $ \case
  True -> Just mempty
  False -> \ma e -> ffor ma $ \a -> Map.singleton a (either (const []) pure e)
```

frontend/src/Frontend/UI/DeploymentSettings.hs
```diff
@@ -572,21 +579,87 @@ uiChainSelection info cls = mdo
...
+  -- above gas payer input
+  divClass "title" $ text "Roles"
+  capabilities <- divClass "group" $ elAttr "table" ("class" =: "table" <> "style" =: "width: 100%") $ case mCaps of
+    Nothing -> el "tbody" $ capabilityInputRows mkSender
+    Just caps -> do
+      el "thead" $ el "tr" $ do
+        elClass "th" "table__heading" $ text "Role"
+        elClass "th" "table__heading" $ text "Capability"
+        elClass "th" "table__heading" $ text "Account"
+      let fixup = fmap (fmap (Map.unionsWith (<>)) . sequence) . sequence
```
alexfmpe: What does the `Maybe` in `Maybe AccountName` mean?

IIUC, the inner `sequence` is 
```
   [Maybe (Map AccountName [SigCapability])] 
-> Maybe [(Map AccountName [SigCapability])]
```
so any `Nothing` in the list will cause the whole thing to be `Nothing`

Why do we want that rather than dropping the failures, e.g. 
```
fmapMaybe id 
:: [Maybe (Map AccountName [SigCapability])] 
-> [(Map AccountName [SigCapability])]
```

tomsmalley: I didn't want to allow submission of a TX by silently dropping partial capabilities

frontend/src/Frontend/UI/DeploymentSettings.hs
tomsmalley: yeah, this could be outdated now. I changed the implementation midway and this might be a leftover

tomsmalley: Actually, I think this is right. Yes, `parseSigCapability` returns `Left` when given no input, but we don't want the error to display for empty inputs.

frontend/src/Frontend/UI/DeploymentSettings.hs
tomsmalley: Resolving based on conversation on mattermost

frontend/src/Frontend/UI/DeploymentSettings.hs
tomsmalley: I feel like `either` is clearer, but this is totally subjective.

mankyKitty (approved):

frontend/src/Frontend/UI/DeploymentSettings.hs
alexfmpe: Maybe rename to `showError` ?

PR #152, opened 2019-10-22
user select for capabilities

mankyKitty:
This allows the user to select an account to use for capabilities in the signing API.

### DIFF
--- a/frontend/src/Frontend/UI/DeploymentSettings.hs
+++ b/frontend/src/Frontend/UI/DeploymentSettings.hs
@@ -687,7 +687,7 @@ uiSenderCapabilities m cid mCaps mkSender = do
   return (firstGASPayer <$> capabilities, capabilities)
   where
     anyGASSigCap :: [SigCapability] -> Bool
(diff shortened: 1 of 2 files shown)


frontend/src/Frontend/UI/DeploymentSettings.hs
```diff
@@ -695,3 +723,23 @@ signingItem (n, _) = do
...
+defaultGASCapability = DappCap
+  { _dappCap_role = "GAS"
+  , _dappCap_description = "Pay the GAS required for this transaction"
+  , _dappCap_cap = PC.SigCapability
+    { PC._scName = PN.QualifiedName
+      { PN._qnQual = PN.ModuleName
+        { PN._mnName = "coin"
+        , PN._mnNamespace = Nothing
+        }
+      , PN._qnName = "GAS"
```
tomsmalley: The capability for gas is `coin.FUND_TX`

mankyKitty: Ta, fixing

PR #157, opened 2019-10-24
Better UX for passphrase recovery / verification.

mankyKitty:
The fields will not be hidden if they have focus, and support using the tab key to
navigate between the fields. The next field is unhidden when it is tab'd to.

Improved the feedback on the copy button for the passphrase creation screen.

### DIFF
--- a/desktop/src/Desktop/Setup.hs
+++ b/desktop/src/Desktop/Setup.hs
@@ -9,10 +9,12 @@
 -- | Wallet setup screens
 module Desktop.Setup (runSetup, form, kadenaWalletLogo) where
 
+import Control.Lens ((<>~))
 import Control.Error (hush)
 import Control.Applicative (liftA2)
 import Control.Monad (unless,void)
 import Control.Monad.IO.Class
+import Data.Bool (bool)
 import Data.Maybe (isNothing, fromMaybe)
 import Data.Bifunctor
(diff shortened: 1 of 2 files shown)


desktop/src/Desktop/Setup.hs
```diff
@@ -281,17 +283,16 @@ passphraseWordElement currentStage k wrd = walletDiv "passphrase-widget-elem-wra
     commonAttrs cls =
       "type" =: "text" <>
       "size" =: "8" <>
-      "class" =: walletClass cls
-
+      "class" =: walletClass cls 
+
```
tomsmalley: whitespace

desktop/src/Desktop/Setup.hs
```diff
@@ -525,7 +530,7 @@ setPassword dSeed = form "" $ do
   pure $ Crypto.generate <$> current dSeed <@> (T.encodeUtf8 <$> pass)
 
   where
-    minPasswordLength = 10
+    minPasswordLength = 1 -- TODO: Reset to realistic min
```
tomsmalley: very secure :smile:

tomsmalley (changes requested):

desktop/src/Desktop/Setup.hs
mankyKitty: dang it! :unlock:

PR #158, opened 2019-10-24
Move _wallet.scss to _setup.scss

mankyKitty:
Original scss did not match conventions, the file and class names have been updated to be
more consistent with their controlling file and their purpose.

### DIFF
--- a/backend/sass/common.blocks/_setup.scss
+++ b/backend/sass/common.blocks/_setup.scss
@@ -1,21 +1,42 @@
 // The wallet ui:
 @import '../include/defs';
 
-.wallet {
-}
-
 $fullscreen-wallet-grey: #f1f1f1;
 $wallet-top-btm-margin: 0.5rem 0 0.5rem 0;
 $kadena-orange: #F3AB3C;
 
-.fullscreen {
+.setup__fullscreen {
   display: grid;
   grid-template-columns: 1fr 1fr 1fr 1fr;
   grid-template-rows: 0.2fr 1.6fr;
   grid-template-areas: "back header header ." ". wrapper wrapper .";
+
+  p {
+    margin: 2rem auto;
+  }
+
+  .checkbox {
+    font-size: 20px;
+    color: #fff;
+    text-align: left;
+    display: inline-block;
+    padding-left: 40px;
+
+    .checkbox__checkmark {
+      top: 2px;
+      height: 20px;
+      width: 20px;
+    }
+    .checkbox__checkmark_type_secondary:after {
+      top: 2px;
+      left: 6px;
+      width: 3px;
+      height: 10px;
+    }
+  }
 }
 
-.fullscreen .wrapper {
+.setup__fullscreen .wrapper {
   grid-area: wrapper;
   text-align: center;
   position: relative;
@@ -30,7 +51,7 @@ $kadena-orange: #F3AB3C;
     background-color: $std-border-color;
     padding: 1rem;
 
-    img.wallet__kadena-logo {
+    img.setup__kadena-logo {
       width: 6rem;
     }
 
@@ -47,7 +68,7 @@ $kadena-orange: #F3AB3C;
   }
 }
 
-.wallet__back {
+.setup__back {
   grid-area: back;
   align-self: stretch;
   padding-top: 2em;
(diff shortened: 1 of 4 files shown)


tomsmalley (approved):

backend/sass/common.blocks/_setup.scss
```diff
@@ -1,23 +1,45 @@
...
   display: grid;
   grid-template-columns: 1fr 1fr 1fr 1fr;
   grid-template-rows: 0.2fr 1.6fr;
   grid-template-areas: "back header header ." ". wrapper wrapper .";
+
+  p {
+    margin: 2rem auto;
+  }
+
+  .checkbox {
```
eskimor, line 16: That's not how BEM is supposed to work. Instead of overriding another block within this block, you would create an element `setup__checkbox`, that is, if you need customizations that are really only relevant in the `setup` block. 

Alternatively, if this is a more general purpose variant of a `checkbox`, you could move that code to the `checkbox` file by means of a modifier class like `checkbox_align_left` or something.

You would also not nest things but have a top level definition:

```
.setup__checkbox {
...
```
in this file or

```
.checkbox_align_left {
....
```

in _checkbox.scss. In HTML you then simply apply all the needed classes, e.g.:

```
elClass "div" "checkbox checkbox_align_left" $ ....
```

By adhering to BEM standards we should actually have maintainable CSS, even in the long run. That's why I moved over to it, from the mess we had before.

A good resource on BEM can be found [here](https://en.bem.info/methodology/quick-start/).

backend/sass/common.blocks/_setup.scss
```diff
@@ -154,94 +176,94 @@ $kadena-orange: #F3AB3C;
...
   }
 }
 
-.fullscreen .wallet__verify-phrase-msg {
+.setup__fullscreen .setup__verify-phrase-msg {
   margin-bottom: 1.5rem;
 }
 
-.fullscreen .wallet__message {
+.setup__fullscreen .setup__message {
```
eskimor: There are some good reasons for nested selectors, but most of the time you would want to avoid them.

eskimor (changes requested):
Ideally we should stick to BEM as much as possible and where it makes sense. I think this would really benefit the project in the short- and more importantly, long run. If I am not the only one with that opinion, we should probably include a note in the README, pointing to appropriate [resources](https://en.bem.info/methodology/quick-start/).

backend/sass/common.blocks/_setup.scss
mankyKitty: I've not encountered BEM before, but I can see the benefits and I'll update this file to suit. Might take me a few tries to get it right. :+1:

eskimor:
Thanks! And sorry about not documenting that better, I thought I had, but as Tom made me realize I only mentioned BEM briefly like 3 times in not perfectly obvious places. So, sorry about that.

mankyKitty:
All good! I appreciate it being pointed out so we're saved future pain.

backend/sass/common.blocks/_setup.scss
```diff
@@ -178,7 +176,7 @@ $kadena-orange: #F3AB3C;
 
 .setup__checkbox-wrapper {
   margin: 1rem;
-  .checkbox {
+  .setup__checkbox {
```
eskimor: Much better - the nesting in `.setup__checkbox-wrapper` should not be necessary though.

backend/sass/common.blocks/_setup.scss
```diff
@@ -250,7 +248,7 @@ $kadena-orange: #F3AB3C;
   margin-bottom: 1.5rem;
 }
 
-.setup__fullscreen .setup__message {
+.setup__password-message {
```
eskimor, line 252: :-)

backend/sass/common.blocks/_setup.scss
```diff
@@ -280,7 +275,7 @@ $kadena-orange: #F3AB3C;
   }
 }
 
-.setup__fullscreen .setup__message-wrapper > .message {
+.setup__message-wrapper > .message {
```
eskimor: Again, we should avoid the nesting. So this should probably simply be `.setup__message`, without any nesting. Same for the `.setup__checkbox` below.

eskimor (changes requested):

PR #160, opened 2019-10-24
Add extraSigners

tomsmalley:


frontend/src/Frontend.hs
```diff
@@ -59,7 +59,7 @@ frontend = Frontend
       , _appCfg_loadEditor = loadEditorFromLocalStorage
       , _appCfg_editorReadOnly = False
       , _appCfg_signingRequest = never
-      , _appCfg_signingResponse = \_ -> pure ()
+      , _appCfg_signingResponse = liftIO . print
```
mankyKitty, line 62: Debugging leftover ?

tomsmalley: Sort of - there's no harm leaving it in, so I figured I may as well. This is never called in the real `frontend` build.

PR #165, opened 2019-10-25
Move wallet and add lock screen

tomsmalley:
The lock screen is currently totally unstyled, but it should work.

Also need to BEMify bits of sass.


frontend/src/Frontend/JsonData.hs
```diff
@@ -306,11 +318,14 @@ makeKeysets walletL cfg =
...
         keys :: Dynamic t (Map KeyName PublicKey)
         keys = do
-          names <- Set.toList <$> keynames
-          let namesMap = Map.fromList . map (,()) $ names
-          wKeys <- walletL ^. wallet_keys
-          let keyPairs = Map.intersection wKeys namesMap
-          pure $ _keyPair_publicKey <$> keyPairs
+          names <- keynames
+          wAccounts <- walletL ^. wallet_accounts
+          -- This isn't great... should really come back to this
```
mankyKitty: Fold over `IntMap.elems wAccounts` ? That would let you do the filtering and building of the new map in a single pass?

frontend/src/Frontend/UI/Wallet.hs
```diff
@@ -122,117 +125,79 @@ uiAvailableKeys aWallet = do
...
   :: MonadWidget t m
-  => (Text, Dynamic t KeyPair)
-  -> m (Event t KeyName)
-uiKeyItem (n, k) = do
-    elClass "tr" "table__row" $ do
-      el "td" $ text n
-
-      let public = keyToText . _keyPair_publicKey <$> k
+  => IntMap.Key
+  -> Dynamic t (SomeAccount key) -- (Text, Dynamic t (KeyPair PrivateKey))
```
mankyKitty: leftover?

frontend/src/Frontend/UI/Wallet.hs
```diff
@@ -122,117 +125,79 @@ uiAvailableKeys aWallet = do
...
 -- distinguish between the two at this point.
-getBalance :: (MonadWidget t m, HasNetwork model t) => model -> ChainId -> Event t AccountName -> m (Event t (Maybe Decimal))
+getBalance
+  :: (MonadWidget t m, HasNetwork model t, HasCrypto key (Performable m))
+  => model -> ChainId -> Event t AccountName -> m (Event t (Maybe AccountBalance))
 getBalance model chain account = do
   networkRequest <- performEvent $ attachWith mkReq (current $ getNetworkNameAndMeta model) account
   response <- performLocalReadCustom (model ^. network) pure networkRequest
-  let toBalance (_, [Right (_, Pact.PLiteral (Pact.LDecimal d))]) = Just d
+  let toBalance (_, [Right (_, Pact.PLiteral (Pact.LDecimal d))]) = Just $ AccountBalance d
```
mankyKitty, line 200: We should probably see if Pact has `Prism`s for this, that could be my preference leaking in though. 

```
toBalance = preview (_2 . _head . _Right . _2 . _PLiteral . _PDecimal. _Just . to AccountBalance)
```

tomsmalley: I don't think they do

frontend/src/Frontend/JsonData.hs
tomsmalley: I converted it to a fold, it's a little neater. I can't remember exactly what I was unhappy about.

PR #166, opened 2019-10-25
Remove read/transact selection from deploy modal.

mankyKitty:


mankyKitty:
Started from the wrong branch

PR #168, opened 2019-10-25
Minor style fixes

tomsmalley:
My macbook really doesn't like all the dependency changes

### DIFF
--- a/frontend/src/Frontend/UI/DeploymentSettings.hs
+++ b/frontend/src/Frontend/UI/DeploymentSettings.hs
@@ -568,7 +568,7 @@ uiSenderDropdown uCfg m chainId = do
            in mkTextAccounts <$> chainId <*> m ^. wallet_accountGuards
   choice <- dropdown Nothing textAccounts $ uCfg
     & dropdownConfig_setValue .~ (if AppCfg.isChainweaverAlpha then never else Nothing <$ updated chainId)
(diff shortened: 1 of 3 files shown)


frontend/src/Frontend/UI/DeploymentSettings.hs
```diff
@@ -621,12 +621,12 @@ capabilityInputRow
...
   -> m (CapabilityInputRow t)
 capabilityInputRow mCap mkSender = elClass "tr" "table__row" $ do
-  (empty, parsed) <- el "td" $ mdo
+  (empty, parsed) <- elClass "td" "table__cell_padded" $ mdo
     cap <- uiInputElement $ def
       & inputElementConfig_initialValue .~ foldMap (renderCompactText . _dappCap_cap) mCap
       & initialAttributes .~
         "placeholder" =: "(module.capability arg1 arg2)" <>
-        "class" =: "input_width_full" <>
+        "class" =: (maybe id (const (<> " input_transparent")) mCap) "input_width_full" <>
```
alexfmpe, line 629:
```suggestion
        "class" =: (bool id (<> " input_transparent") (isJust mCap)) "input_width_full" <>
```

tomsmalley: I don't like `bool` because I have to open hoogle every single time I see it

PR #169, opened 2019-10-25
Fix warnings

alexfmpe:
I get a lot of
```
src/Desktop/Setup.hs:118:6: warning: [-Wsimplifiable-class-constraints]
    • The constraint ‘MonadWidget t m’ matches an instance declaration
      instance Reflex.Dom.Old.MonadWidgetConstraints t m =>
               MonadWidget t m
        -- Defined in ‘Reflex.Dom.Old’
      This makes type inference for inner bindings fragile;
        either use MonoLocalBinds, or simplify it using the instance
    • In the type signature:
        createNewWallet :: MonadWidget t m =>
                           Maybe (Crypto.MnemonicSentence 15) -> SetupWF t m
```
only in this module for some reason. Rather than figure it out, I just used more granular constraints

### DIFF
--- a/desktop/src/Desktop/Mac.hs
+++ b/desktop/src/Desktop/Mac.hs
@@ -13,19 +13,16 @@ import Control.Monad (forever)
 import Control.Monad.Except (throwError)
 import Control.Monad.IO.Class
 import Data.Foldable (for_)
-import Data.Proxy (Proxy(..))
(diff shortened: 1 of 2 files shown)


mankyKitty:
Bonus weird points because I didn't encounter any of those warnings...

PR #172, opened 2019-10-28
Add 'kadena address'

mankyKitty:
Adding the types and underlying functionality for a 'kadena account'.


frontend/src/Frontend/KadenaAddress.hs
```diff
@@ -0,0 +1,243 @@
...
+
+parseKadenaAddressBlob :: A.Parser (AccountName, ChainId, NetworkName, Word32)
+parseKadenaAddressBlob = (,,,)
+  <$> parseAccountName
+  <*> parseChainId
+  <*> parseNetworkName
+  <*> A.hexadecimal
+  where
+    failParse = either fail pure
+    mkParser p c = (c <$> p) >>= failParse
```
tomsmalley: I needed to add a type signature here to get it to compile

frontend/src/Frontend/KadenaAddress.hs
```diff
@@ -0,0 +1,243 @@
...
+  , _kadenaAddress_checksum :: Word32
+    -- ^ The CRC32 checksum of the AccountName, ChainId, and NetworkName
+  }
+  deriving (Show, Eq)
+
+textKadenaAddress :: KadenaAddress -> Text
+textKadenaAddress = TE.decodeUtf8With TE.lenientDecode . _kadenaAddress_encoded
+
+humanReadableDelimiter :: Word8
+humanReadableDelimiter = 93 -- ']'
```
tomsmalley, line 123: I'd have a minor preference for `|`. No reason other than it looks better :smile:

tomsmalley (dismissed):
Looks good. We should also restrict AccountName to latin1.

frontend/src/Frontend/KadenaAddress.hs
tomsmalley: Come to think of it, account names can have this character.

tomsmalley (changes requested):
We can't have the account name encoded as latin1 and also use a latin1 delimiter without doing some form of escaping.

frontend/src/Frontend/KadenaAddress.hs
mankyKitty: Account names can currently have everything in latin1, so this was a placeholder at best. :< Also I agree that `|` looks better. I'll make a ticket.

tomsmalley:
Merging since we have tickets to address the delimiter concerns

PR #181, opened 2019-10-31
Create 'add account' dialog

mankyKitty:
This PR adds the dialog for creating a wallet only account, with the inactive button for creating vanity accounts.


frontend/src/Frontend/UI/Dialogs/AddAccount.hs
```diff
@@ -0,0 +1,93 @@
...
+        value <$> mkLabeledClsInput inpElem "Notes"
+
+      _ <- accordionItem' False "add-account__advanced-content" (text "Advanced") $
+        confirmButton (def & uiButtonCfg_disabled .~ pure True) "Create Vanity Account"
+
+      pure (dChainId, dNotes)
+
+  modalFooter $ do
+    onCancel <- cancelButton def "Cancel"
+    text " "
```
alexfmpe: Placeholder for margin?

frontend/src/Frontend/UI/Dialogs/AddAccount.hs
```diff
@@ -0,0 +1,93 @@
...
+
+uiWalletOnlyAccountCreated
+  :: forall t m mConf
+     . ( MonadWidget t m
+       , Monoid mConf
+       )
+  => AccountName
+  -> Event t ()
+  -> m (mConf, Event t ())
+uiWalletOnlyAccountCreated newAccount _onClose = do
```
alexfmpe: Is `_onClose` placeholder or just making the types agree with `ModalImpl` ?

frontend/src/Frontend/UI/Dialogs/AddAccount.hs
```diff
@@ -0,0 +1,93 @@
...
+  _ <- modalMain $ divClass "segment modal__main wallet_only__account-created-modal" $ do
+    elClass "h2" "heading heading_type_h2" $
+      text "[WALLET IMAGE PLACEHOLDER]"
+    divClass "wallet-only__account-created-details" $ do
+      divClass "wallet-only__account-heading" $ text "Account Created:"
+      divClass "wallet-only__account-name" $ text (unAccountName newAccount)
+
+  modalFooter $ do
+    onConfirm <- confirmButton def "Done"
+    pure $ (mempty, leftmost [onClose, onConfirm])
```
alexfmpe: It seems like this function doesn't actually care at all about `mconf`. Is the `mempty` a placeholder?

mankyKitty: Both this and the `onClose` are part of the `ModalImpl` types.

frontend/src/Frontend/UI/Dialogs/AddAccount.hs
mankyKitty: Copy-pasta from the modal I was building my implementation from, I suspect. But yes, a placeholder for a margin that should be done in CSS. I'll remove it.

frontend/src/Frontend/UI/Dialogs/AddAccount.hs
alexfmpe: Yeah I'm just wondering whether the type-massaging belongs in this function or in the caller

frontend/src/Frontend/UI/Dialogs/AddAccount.hs
tomsmalley: type alignment

frontend/src/Frontend/ReplGhcjs.hs
```diff
@@ -93,16 +93,16 @@ app sidebarExtra appCfg = void . mfix $ \ cfg -> do
       mkPage r c = elDynAttr "div" (ffor (demuxed route r) $ \s -> "class" =: (c <> if s then " page__content visible" else " page__content"))
   updates <- divClass "page" $ do
     netCfg <- networkBar ideL
-    walletCfg' <- mkPage (FrontendRoute_Wallet :/ ()) "wallet" $ do
-      uiWallet (_ide_wallet ideL)
+    (walletCfg', mconf) <- mkPage (FrontendRoute_Wallet :/ ()) "wallet" $ do
+      uiWallet (_ide_wallet ideL) ideL
```
tomsmalley: No point in the first argument now

frontend/src/Frontend/UI/Dialogs/AddAccount.hs
tomsmalley: Since this is pre-modal redesign, perhaps we should remove all uses of this and let the redesign work fix it in CSS

frontend/src/Frontend/UI/Wallet.hs
```diff
@@ -70,14 +71,25 @@ type HasUiWalletModelCfg mConf key m t =
...
-    onCreate <- uiCreateKey w
-    keysCfg <- uiAvailableKeys w
-
-    pure $ keysCfg & walletCfg_genKey .~ fmap (\a -> (a, "0", "")) onCreate -- TODO let user pick chain/notes
+  -> ModalIde m key t
+  -> m (mConf, ModalIdeCfg m key t)
+uiWallet w m = divClass "keys group" $ do
+  eOpenAddAccount <- confirmButton def "+ Add Account"
+
+  onCreate <- uiCreateKey w
```
tomsmalley, line 83: Seems like we no longer need this?

frontend/src/Frontend/UI/Wallet.hs
```diff
@@ -70,14 +71,25 @@ type HasUiWalletModelCfg mConf key m t =
...
+  keysCfg <- uiAvailableKeys w
+
+  pure ( keysCfg & walletCfg_genKey .~ fmap (\a -> (a, "0", "")) onCreate -- TODO let user pick chain/notes
+       , mempty & modalCfg_setModal .~ leftmost
+         [ Just (uiCreateWalletOnlyAccount m) <$ eOpenAddAccount
+         , Just . uiWalletOnlyAccountCreated <$> (w ^. wallet_walletOnlyAccountCreated)
```
tomsmalley: The presence of two dialogs here seems off, I think we should just have one which contains a workflow.

tomsmalley (changes requested):
~Does this modal not have to send a TX to create the account?~

frontend/src/Frontend/UI/Dialogs/AddAccount.hs
mankyKitty: Removed `text " "` and added a ticket to fix this with the correct CSS.

frontend/src/Frontend/UI/Dialogs/AddAccount.hs
```diff
@@ -7,34 +8,53 @@ module Frontend.UI.Dialogs.AddAccount
...
+       , HasAddAccountModelCfg model mConf key t
        )
-  => model
+  => ModalIde m key t
   -> Event t ()
   -> m (mConf, Event t ())
-uiCreateWalletOnlyAccount model _onClose = do
-  onClose <- modalHeader $ text "Add Account"
+uiCreateWalletOnlyAccount model onClose = mdo
+  onClose <- modalHeader $ dynText title
```
alexfmpe, line 50: Shadows argument

mankyKitty: Something very weird happened here... The code you commented on did indeed shadow that variable. However my local version did not as the input to the function was named `_onClose` not `onClose`.

Updating this code to rename the variable to `onCloseExternal` now all of a sudden shows the input to the function as having been  `_onClose` and not `onClose`. :/

frontend/src/Frontend/UI/Dialogs/AddAccount.hs
```diff
@@ -0,0 +1,121 @@
...
+  , HasWalletCfg mConf key t
+  )
+
+uiAddWalletOnlyAccountDialogButton
+  :: forall t m model key mConf
+     . ( MonadWidget t m
+       , HasAddAccountModelCfg model mConf key t
+       )
+  => ModalIde m key t
+  -> m (ModalIdeCfg m key t)
```
tomsmalley: This should be fully polymorphic

tomsmalley (changes requested):
I wired this into the wallet UI for testing, but I get a causality loop exception when finishing the workflow after creating the account.

PR #182, opened 2019-10-31
Fix routes

tomsmalley:


frontend/src/Frontend/Storage.hs
```diff
@@ -115,6 +115,9 @@ class HasStorage m where
 instance Monad m => HasStorage (StorageT m) where
   askStorage = StorageT ask
 
+instance (HasStorage m, Monad m) => HasStorage (RoutedT t r m) where
+  askStorage = lift askStorage
```
alexfmpe: https://github.com/kadena-io/chainweaver/blob/c5b763b15500351292c910ff58291c928ebf1ee1/frontend/src/Frontend/Storage.hs#L112

frontend/src/Frontend/ReplGhcjs.hs
```diff
@@ -142,7 +150,7 @@ walletSidebar sidebarExtra = elAttr "div" ("class" =: "sidebar") $ do
 -- | Get the routes to the icon assets for each route
 routeIcon :: R FrontendRoute -> (Text, Text)
 routeIcon = \case
-  FrontendRoute_Main :/ () -> (static @"img/menu/contracts.png", static @"img/menu/contracts_highlighted.png")
+  FrontendRoute_Contracts :/ Nothing -> (static @"img/menu/contracts.png", static @"img/menu/contracts_highlighted.png")
```
alexfmpe: Doesn't this yield a warning? We're not handling `FrontendRoute_Contracts :/ Just r`

frontend/src/Frontend/Storage.hs
tomsmalley: hehe

frontend/src/Frontend/ReplGhcjs.hs
tomsmalley: There's a wildcard further down, but I just stuck wildcards after `:/`, seems to make more sense

PR #188, opened 2019-11-05
Deleting password input now prevents wallet setup progress

mankyKitty:
Deleting the password field after entering a password that passes validation
will now cause the validation to revert to a failed state with the appropriate
error message.

### DIFF
--- a/desktop/src/Desktop/Setup.hs
+++ b/desktop/src/Desktop/Setup.hs
@@ -143,7 +143,7 @@ walletSetupRecoverHeader currentScreen = setupDiv "workflow-header" $ do
       else
         []
 
-    line cls sid = 
+    line cls sid =
       elClass "div" (T.unwords $ setupClass "workflow-icon": setupClass "header-line" : cls : addActive sid) blank
 
     isActive sid = sid `elem` (progress currentScreen)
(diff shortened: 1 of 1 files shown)


desktop/src/Desktop/Setup.hs
```diff
@@ -519,20 +518,20 @@ setPassword dSeed = form "" $ do
...
-  let p1 = current $ value p1elem
-      p2 = current $ value p2elem
+  p1Dirty <- holdUniqDyn =<< holdDyn False (True <$ _inputElement_input p1elem)
+  p2Dirty <- holdUniqDyn =<< holdDyn False (True <$ _inputElement_input p2elem)
 
-      inputsNotEmpty = not <$> liftA2 (||) (T.null <$> p1) (T.null <$> p2)
+  let inputsDirty = current $ liftA2 (&&) p1Dirty p2Dirty
 
-  eCheckPassword <- fmap (gate inputsNotEmpty) $ delay 0.4 $ leftmost
+  eCheckPassword <- fmap (gate inputsDirty) $ delay 0.2 $ leftmost
```
alexfmpe, line 528: If I'm reading this right, errors won't be shown until both fields have received input - is that intended?

mankyKitty: Yes that is intentional, otherwise the errors for password length are displayed as soon as the user lands on the screen.

mankyKitty: Changed the `&&` to `||` so the dirty check is faster to pass, gives snappier feedback for password validation. :) :+1:

PR #191, opened 2019-11-05
Redesign wallet

alexfmpe:


backend/sass/common.blocks/_wallet.scss
```diff
@@ -0,0 +1,44 @@
...
+   For more info: https://en.bem.info/methodology/quick-start/
+
+*********************************************************************************/
+
+@import '../include/defs';
+
+.keys-list {
+  margin-top: 20px;
+  margin-left: 22px;
+}
```
eskimor, line 12: This should be renamed to for example `wallet__keys-list`.

PR #192, opened 2019-11-05
Add some convenience scripts

alexfmpe:


devDesktop
```diff
@@ -1 +1 @@
-ob run --import desktop:Desktop.Frontend --frontend Desktop.Frontend.desktop
+ob run --import desktop:Desktop --frontend Desktop.desktop
```
alexfmpe: This way `Desktop.Mac` is also typechecked

mankyKitty:
Should we put these in a `scripts` folder in the root of the project? Just keep things organised.

PR #195, opened 2019-11-06
Fix account balance preview

tomsmalley:


common/src/Common/Network.hs
```diff
@@ -208,3 +226,82 @@ renderNodeRef (NodeRef (URI.Authority mUser h mp)) =
...
+        t -> Left $ "Unexpected PactValue (expected decimal): " <> renderCompactText t
+      acc <- mkAccountName accountText
+      pure (acc, bal)
+    pure $ Map.fromList m
+  v -> Left $ "Unexpected PactValue (expected object): " <> renderCompactText v
+
+-- | Get the last result, as we would under unwrapped deployment.
+parseResults :: PactValue -> Either Text PactValue
+parseResults = first ("parseResults: " <>) . \case
+  PList vec -> pure $ V.last vec
```
mankyKitty: `V.last` is partial, if we have an empty list this will throw an exception. Maybe...

```haskell
vec ^? _last
```

common/src/Common/Network.hs
```diff
@@ -208,3 +226,82 @@ renderNodeRef (NodeRef (URI.Authority mUser h mp)) =
...
+      { _tObject = Pact.Types.Term.Object
+        { _oObject = ObjectMap $ Map.fromList $ map getBalance $ Set.toAscList accounts
+        , _oObjectType = TyPrim TyDecimal
+        , _oKeyOrder = Nothing
+        , _oInfo = def
+        }
+      , _tInfo = def
+      }
+    -- It would be nice to parse and compile the code and shove it into a
+    -- giant 'Term' so we can serialise it, but 'pretty' is not guaranteed to
```
mankyKitty, line 292: Not an issue for us/this release, but should probably pass this back to Kadena as it'd be nice to have some TH like `[pact| (foo "biscuit")|]` that would give us compile errors if given an invalid pact expression.

common/src/Common/Network.hs
tomsmalley: the joys of partial functions which aren't marked `unsafe` :(

PR #204, opened 2019-11-08
Change kadena address delimiter to '|'

mankyKitty:


frontend/src/Frontend/KadenaAddress.hs
```diff
@@ -120,7 +120,7 @@ textKadenaAddress :: KadenaAddress -> Text
 textKadenaAddress = TE.decodeUtf8With TE.lenientDecode . _kadenaAddress_encoded
 
 humanReadableDelimiter :: Word8
-humanReadableDelimiter = 93 -- ']'
+humanReadableDelimiter = 124 -- '|'
```
alexfmpe: Should we use [ord](https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Char.html#v:ord) here?

mankyKitty: Yeah we can, I used `ord` to get the 124 value anyway. But using it directly might be a bit clearer and I suspect the compiler would inline it away anyway.

alexfmpe: Also found this https://hackage.haskell.org/package/bytestring-0.10.10.0/docs/Data-ByteString-Internal.html#v:c2w

mankyKitty: Haha its implementation is what I just added. Docs might be better if I use that function because then someone has to look it up, thoughts ?

alexfmpe: On the fence - `ord` should be more well-known though.

PR #209, opened 2019-11-12
Only display accounts from current network

tomsmalley:
* Add network into account type
* Filter the accounts we display in the wallet based on the currently
  selected network

### DIFF
--- a/frontend/src/Frontend/UI/Wallet.hs
+++ b/frontend/src/Frontend/UI/Wallet.hs
@@ -201,7 +201,6 @@ uiKeyItem selectedNetwork i d = do
 
             pure $ leftmost
               [ mkDialog AccountDialog_Details onDetails
-  --            , mkDialog AccountDialog_Details onDetails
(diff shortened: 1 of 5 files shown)


frontend/src/Frontend/UI/Wallet.hs
```diff
@@ -170,44 +158,51 @@ uiKeyItems aWallet aNetwork = do
...
+
+            void $ receiveButton $ cfg & uiButtonCfg_disabled .~ True
+            void $ sendButton $ cfg & uiButtonCfg_disabled .~ True
+            onDetails <- detailsButton cfg
+
+            let mkDialog dia onE = (\a -> (dia, i, a)) <$> current account <@ onE
+
+            pure $ leftmost
+              [ mkDialog AccountDialog_Details onDetails
+  --            , mkDialog AccountDialog_Details onDetails
```
mankyKitty: Leftover?

PR #215, opened 2019-11-14
feat: Adds the resources page

benkolera:
The asana ticket just had a screenshot, so I'm not entirely certain what these buttons need to be linking to. But now that I've done the CSS dance it should be easy to get the links right from this, so just lmk what they should be. 

In the user story:

Resources

My UI:

Which is not pixel perfect the same yet, but given that the user story didn't have a launch icon on the documentation, I'm probably missing something important and don't want to make things pixel perfect till I have all the links clear in my head. 

I went ahead and assumed that we wanted the whole grid clickable rather than just the box. It does mean that you can click anywhere in a big poorly visually distinguished squares which is a little weird. Can make two separate hrefs for the icon and title box if that's weird: because someone for sure is going to click the icon. :slightly_smiling_face:

### DIFF
--- a/frontend/src/Frontend/ReplGhcjs.hs
+++ b/frontend/src/Frontend/ReplGhcjs.hs
@@ -109,8 +109,11 @@ app sidebarExtra appCfg = void . mfix $ \ cfg -> do
           envCfg <- rightTabBar "main__right-pane" ideL
           pure $ uiEditorCfg <> envCfg
         pure $ controlCfg <> mainCfg
-      -- TODO resources page
-      FrontendRoute_Resources -> text "Resources" >> pure mempty
+      FrontendRoute_Resources -> mkPageContent "resources" $ do
+        controlBar "Resources" blank
+        elClass "main" "main page__main" $ do
+          resourcesWidget
+        pure mempty
       -- TODO settings page
       FrontendRoute_Settings -> text "Settings" >> pure mempty
     flattenedCfg <- flatten =<< tagOnPostBuild routedCfg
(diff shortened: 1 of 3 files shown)


frontend/src/Frontend/ReplGhcjs.hs
```diff
@@ -348,3 +351,23 @@ headerBtnCfg
...
+  resourceCell "Support" (static @"img/resources/support.svg") "#"
+    "Explore Help Resources to learn about Chainweaver, solve problems and get in touch"
+  resourceCell "Documentation" (static @"img/resources/documentation.svg") "http://pact-language.readthedocs.io/"
+    "Complete technical resources for Chainweaver, Kadena blockchain and Pact language"
+  resourceCell "Tutorials" (static @"img/resources/tutorials.svg") "https://pactlang.org/"
+    "Read or watch tutorials for writing smart contracts using the Pact language"
+  where
+    resourceCell title iconUrl href desc =
+      elAttr "a" ("class" =: "resources__cell" <> "href" =: href <> "target" =: "_blank")  $ do
+        elAttr "div" ("class" =: "resources__cell-icon" <> "style" =: ("background-image: url(" <> iconUrl <>")")) $ blank
```
benkolera, line 368: Do we ever use the impure static obelisk urls (i.e the ones without the hash in the uri that redirect) as links in our sass files? It's a bit weird putting this out in an inline style just because we need to be in haskell to use the static function.

benkolera: Oh yeah, we totally do in fonts. Is that preferred over the inline style here?

benkolera:
From standup today, it sounded like we don't actually have firm links for the resources yet. This looks like a good enough progress until we have that. Feel free to review as a non-wip PR and be as savage as you like. :slightly_smiling_face:

frontend/src/Frontend/ReplGhcjs.hs
alexfmpe: > It's a bit weird putting this out in an inline style just because we need to be in haskell to use the static function.

AFAICT there are two ways to do this in haskell without inline style. 
You can use classes here and then either throw the CSS in a `<style>` in the [`<head>`](https://github.com/kadena-io/chainweaver/pull/197/commits/5e0475dcfb2ab90adc9656ba149f69ad8e64ccec#diff-9f2ed08cd073fc795038ebf78b9b0137R104) tag or use some EDSL CSS preprocessor. 
Personally I'm not opposed to inline style but for consistency I'd recommend the `<head>` route.

PR #216, opened 2019-11-14
fix: Adds the missing launch icon to the help button

benkolera:
The existing launch svg was not a nice color (it's #D7D7D7, which is the hover background color of the button, lol) to use directly on a default colored button, so I made a new launch svg that is the same color as receive.svg so that it works better with the hover and active states of the button.


desktop/src/Desktop/Frontend.hs
```diff
@@ -1,14 +1,14 @@
-{-# LANGUAGE ConstraintKinds #-}
```
benkolera: It looked like we used stylish haskell in a different file that I changed earlier today, but this one did not have the autoformat previously applied. Should I have this autoformat turned on or off working on chainweaver?

desktop/src/Desktop/Frontend.hs
```diff
@@ -184,7 +184,9 @@ lockScreen xprv = setupDiv "fullscreen" $ divClass "wrapper" $ setupDiv "splash"
 
     e <- confirmButton (def & uiButtonCfg_type ?~ "submit") "Unlock"
     (_help, restore) <- setupDiv "button-horizontal-group" $ do
-      help <- uiButton def $ text "Help" -- TODO where does this go?
+      help <- uiButton def $ do
+        elAttr "img" ("src" =: static @"img/launch_dark.svg" <> "class" =: "button__text-icon") blank
+        text "Help" -- TODO where does this go?
```
benkolera, line 189: It's worth noting that the help button is still a placebo button. :slightly_smiling_face:

desktop/src/Desktop/Frontend.hs
tomsmalley: off please; the diffs are bad otherwise

benkolera: Fixed, as well as adding in https://github.com/kadena-io/chainweaver/pull/223 to get consistency around the place.

PR #218, opened 2019-11-15
feat: Adds settings page and removes network cog from toolbars

benkolera:
The idea is that we have some kind of settings page that we can turn
items on/off based on what type off app we are running. Right now, we
only have then network config, and we don't have a non-modal design
so this is just a basic first wave where we:

- Only have network
- Network is still a modal

It's expected that this will need some more work once we have designs for
a non-modal network form and also once we have other settings, but this
feels like a good start.

Once https://github.com/kadena-io/chainweaver/pull/215 is merged, there probably should be some refactoring to commonise those icony button thingos. :slightly_smiling_face:


frontend/src/Frontend/AppCfg.hs
```diff
@@ -6,6 +6,10 @@ import Language.Javascript.JSaddle (JSM)
 import Reflex.Dom
 import Kadena.SigningApi
 
+data EnabledSettings = EnabledSettings
+  { _enabledSettings_network :: Bool
```
tomsmalley: Not sure I understand why we have this type?

benkolera: The asana ticket mentions that we need some way of configuring turning settings on and off depending on what distro (app vs web) chainweaver is running in. The idea would be that they'd be a bool per setting and each app would turn it off or on. It's hard to really think about till we have the other settings, but it was moving in that direction. 

> There needs to be a way of adding more icons and pages depending on if we are in the web app or desktop app.

tomsmalley: Ahh I see. In the case of the network settings, we don't need this: it's enabled everywhere. The settings I was referring to there were the desktop specific bits. Can you remove this field and leave the settings as a stub? (Or remove them entirely, but I feel we will actually need something like this soon).

PR #222, opened 2019-11-18
feat: Expands sidebar on hover

benkolera:
On hover, the sidebar will now slide out over the page content with a 200ms animation showing some alt text so that we don't just have icons. Design was such that the top left icon stays put, so it's a little bit funkier than usual. 

feature screenshare


tomsmalley:
Nice

PR #223, opened 2019-11-18
refactor: Unstylish the language pragmas

benkolera:
This ran stylish-haskell over every file ordering pragmas but unaligning their curly braces. Two other code fixes to get stylish haskell to parse the codebase.


tomsmalley:
Merging before we have to create another PR for this :smile:

PR #232, opened 2019-11-19
fix: Centers the wallet image on the AddAccount success screen

benkolera:
Prior to this we had two images and absolutely positioning one over the other.
This meant moving the base meant twiddling pixels with the other. This now just
uses a block with a background that should be much more robust. The only thing
to be careful of here is the relationship between the height of the div and the
vertical positioning of the inner image, which is still better than before. :)

### DIFF
--- a/backend/sass/common.blocks/_modal.scss
+++ b/backend/sass/common.blocks/_modal.scss
@@ -216,7 +216,6 @@
 }
 
 .wallet_only__account-created-done-splash-bg {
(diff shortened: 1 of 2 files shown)


benkolera:
Oh, actually because there are no css transforms on background images, we transform the div and have to transform the inner image to unflip it... That's pretty gross. Suggest just mirroring the image if we hate that. :)

PR #239, opened 2019-11-20
A bunch of design tweaks

benkolera:
Look at the individual commits for indvidual change sets. I don't think that any of them are controversial and I'm sure people don't want 10+ prs. I can revert commits if we hate specific ones.

The basic gist:
- Went to fix the poor quality of the wallet icon on the setup splash screen. Did this with the svg version rather than the png.
- Played around with sizings on the wallet list, made the network ok signal chainweaver blue. Made the add account button the right colour.
- On the pre-passphrase warning screen, the first connecting line should be highlighted.
- Made the text actually the full width on the passphrase warning screen.
- Refactored the splash wallet logo screens to be the background image variety.
- Removed skip button on confirm passphrase for release

There are still some obvious differerences between zeplin and our code. Namely some of the dialogs are the old design: 

In chainweaver:
in chainweaver

vs Zeplin:
zeplin

Most of my focus has been on the wallet list and setup. Haven't gone into the ide too much yet. 

I'm annoyed at how I can't even make up my mind about absolute vs relative sizing. I tend to side with relative sizing, but given that we probably don't care about font scaling too much and we are trying to line things up with zeplin, may as well go with px.


desktop/src/Desktop/Setup.hs
```diff
@@ -533,13 +533,10 @@ confirmPhrase eBack dPassword mnemonicSentence = Workflow $ do
...
   let done = (== actualMap) <$> dConfirmPhrase
 
-  -- TODO: Remove me before release, I'm a dev hack
-  skip <- uiButton btnCfgTertiary $ text "Skip"
-
   continue <- continueButton (fmap not done)
 
   finishSetupWF WalletScreen_VerifyPassphrase $ leftmost
-    [ doneScreen <$> tagMaybe (current dPassword) (leftmost [continue, skip])
+    [ doneScreen <$> tagMaybe (current dPassword) (leftmost [continue])
```
alexfmpe: Weird leftmost but ok

benkolera: Oh yeah, that's a good point. I'll fix that.

PR #245, opened 2019-11-20
fix: Gives the wallet table some breathing room.

benkolera:
Was forcing a scrollbar, and shouldn't be up against the side of the page anyway.

### DIFF
--- a/backend/sass/common.blocks/_wallet.scss
+++ b/backend/sass/common.blocks/_wallet.scss
@@ -19,11 +19,18 @@
 
 .wallet__table-heading {
   @extend .table__heading;
-  padding: 15px 30px 30px 15px;
+  padding: 15px;
+  padding-bottom: 30px;
 }
 
 .wallet__table-cell {
-  padding: 15px 30px 15px 15px;
+  padding: 15px;
+  text-overflow: ellipsis;
+  overflow: hidden;
+}
+
+.wallet__table-wallet-address {
+  width: 15em;
   text-overflow: ellipsis;
   overflow: hidden;
 }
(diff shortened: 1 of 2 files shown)


frontend/src/Frontend/UI/Wallet.hs
```diff
@@ -127,16 +127,16 @@ uiKeyItems model = do
     keyMap' = model ^. wallet_accounts
     keyMap = Map.fromAscList . IntMap.toAscList <$> keyMap'
     tableAttrs =
-      "style" =: "table-layout: fixed; width: 100%"
+      "style" =: "table-layout: fixed; width: 98%"
```
alexfmpe, line 130: How about leaving it at 100% and use a fixed margin?

backend/sass/common.blocks/_wallet.scss
```diff
@@ -19,12 +19,11 @@
...
 .wallet__table-heading {
   @extend .table__heading;
-  padding: $medium-padding;
-  padding-bottom: $bigger-padding;
+  padding: 15px 30px 30px 15px;
 }
 
 .wallet__table-cell {
-  padding: $medium-padding;
+  padding: 15px 30px 15px 15px;
```
alexfmpe: Right padding is greater than left padding?

frontend/src/Frontend/UI/Wallet.hs
benkolera: Something about it being at 100% was causing a scrollbar. And it's too late to figure out exactly why. It's more of a workaround to get rid of the scrollbar than just a margin.

backend/sass/common.blocks/_wallet.scss
benkolera: Yeah. The rejected ticket was explicitly mentioning the fact that there wasn't enough padding between the keys. This way being lazy to get that effect, but you are right that it is confusing and I've done it a better way in 83dce35

PR #250, opened 2019-11-21
Only add empty capability when capabilities aren't specified

alexfmpe:


alexfmpe:
Wrong fix - see #255 instead

PR #251, opened 2019-11-21
Make namespaces significantly less broken

alexfmpe:
This seems like it should fix it, and it does make "Load into REPL" work the first time, but it errors after that one with
```
<interactive>:47:0: : Failure: Database exception: Insert: value already at key: free.hello-world_message
```
Also shows a warning on the editor unless `(namespace "free")` is moved to the end:
```
1:1 Unable to resolve module hello-world
```


alexfmpe:
The `Insert: value already at key:` error is actually correct behavior since 'Load into REPL' no longer resets the state.

PR #254, opened 2019-11-21
Kadena Address Changes

mankyKitty:
Change the checksum to pure haskell CRC32 using `crc` package.

- Remove network name from the address.
- Add a flag to indicate if the account has been created yet, defaults to no.
- Remove `uri-bytestring` dependency.
- Update test.

### DIFF
--- a/frontend/frontend.cabal
+++ b/frontend/frontend.cabal
@@ -79,7 +79,7 @@ library
                , unordered-containers
                , kadena-signing-api
                , base64-bytestring
-               , uri-bytestring
+               , crc
                , attoparsec
                , vector
 
--- a/frontend/frontend.cabal
+++ b/frontend/frontend.cabal
@@ -38,7 +38,6 @@ library
                , free
                , generic-deriving
                , ghcjs-dom
-               , hashing
                , http-client
                , http-client-tls
                , http-types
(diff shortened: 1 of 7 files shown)


frontend/frontend.cabal
```diff
@@ -79,7 +79,7 @@ library
                , unordered-containers
                , kadena-signing-api
                , base64-bytestring
-               , uri-bytestring
```
alexfmpe, line 82: Can probably remove `hashing` too

frontend/src/Frontend/KadenaAddress.hs
```diff
@@ -101,21 +102,31 @@ data KadenaAddressError
   | InvalidHumanReadablePiece Text
   deriving (Show, Eq)
 
+data KadenaAddressPrefix
+  = KadenaAddressPrefix_Keep
+  | KadenaAddressPrefix_None
+  deriving (Show, Eq)
+
+data KadenaAddress_AccountCreated
```
alexfmpe: Don't think we actually use `_` outside of field/constructor prefixes, convenient as it would be. Might mess with lens-making?

frontend/src/Frontend/KadenaAddress.hs
```diff
@@ -131,57 +141,65 @@ bsshow = fromString . show
...
-      acc
-      cid
-      network
-      checksum
+    name
+    <> cons humanReadableDelimiter cid
+    <> cons humanReadableDelimiter (case _kadenaAddress_accountCreated ka of
+                                      AccountCreated_Yes -> encoded
+                                      AccountCreated_No -> checksum
+                                   )
```
alexfmpe:
```haskell
intersperse humanReadableDelimiter
[ name
, cid
, case _kadenaAddress_accountCreated ka of
    AccountCreated_Yes -> encoded
    AccountCreated_No -> checksum
]

PR #258, opened 2019-11-22
feat: Returns all failover errors from pact calls

benkolera:
We have a request to present all of the errors that happen when we get an error that looks to be transient to a node (i.e we should failover to another node). Otherwise those errors just disappear, and there may be something helpful to the user in there.


frontend/src/Frontend/Network.hs
```diff
@@ -762,35 +787,32 @@ performNetworkRequestCustom
...
-            [] -> doReq Nothing req
-
+                               then go (errs <> DList.singleton err) ns
+                               else pure $ mkResult errs (Left err)
+                Right res -> pure $ mkResult errs (Right res)
+            [] -> mkResult errs <$> doReq Nothing req
+          mkResult errs (Left e) = This (NEL.fromList $ DList.toList (errs <> pure e))
+          mkResult errs (Right res) = maybe (That res) (flip These res) . nonEmpty . DList.toList $ errs
+
+    doReq :: Maybe NodeInfo        -> NetworkRequest -> JSM (Either NetworkError (Maybe Gas, PactValue))
```
tomsmalley: strange spacing

frontend/src/Frontend/Network.hs
```diff
@@ -762,35 +787,32 @@ performNetworkRequestCustom
...
-                               then go ns
-                               else pure errRes
-                Right _ -> pure errRes
-            [] -> doReq Nothing req
-
+                               then go (errs <> DList.singleton err) ns
+                               else pure $ mkResult errs (Left err)
+                Right res -> pure $ mkResult errs (Right res)
+            [] -> mkResult errs <$> doReq Nothing req
+          mkResult errs (Left e) = This (NEL.fromList $ DList.toList (errs <> pure e))
```
tomsmalley: I'm not convinced of the value of using DList given the length of the list is always going to be fairly short. I'd rather just keep the NonEmpty list and either just pay the append cost or cons everything and reverse it at the end. We can avoid the NEL.fromList that way too.

tomsmalley (commented):
Looks good; just nitpicks

frontend/src/Frontend/Network.hs
benkolera: Yeah, that's fair. It's just a reflex when doing this sort of stuff. :)

frontend/src/Frontend/Network.hs
benkolera: Yup, not sure what happened there. Lol.

frontend/src/Frontend/Network.hs
benkolera: Fixed by d672530

frontend/src/Frontend/Network.hs
benkolera: Fixed by d672530

benkolera:
OK, should be good for proper review now. Was tricky to test because I'm a noob and kept getting distracted with other things.

PR #236, opened 2019-11-19
Move preview to tab pane instead of workflow page

alexfmpe:

PR #260, opened 2019-11-22
fix: Adds text-wrap to preview tab raw response.

benkolera:
When the preview tab was created in https://github.com/kadena-io/chainweaver/pull/236/files#diff-55712e3137bf1a0aefa5463d98924d80R1004
there was UI copied from DeployConfirmation. I fixed the UI in DeployConfirmation
but that UI got copied before the fix and I didn't pick this up in review, sorry!


tomsmalley:
It would appear none of us picked it up. Thanks!

PR #261, opened 2019-11-22
Transfer create

tomsmalley:


frontend/src/Frontend/UI/Dialogs/Send.hs
```diff
@@ -118,12 +123,17 @@ sendDeploy
...
-        , tshow amount
-        , ")"
+  let code = T.unwords $ catMaybes
+        [ Just $ "(coin." <> case _kadenaAddress_accountCreated recipient of
+          AccountCreated_Yes -> "transfer"
+          AccountCreated_No -> "transfer-create"
+        , Just $ tshow $ unAccountName $ _account_name sender
+        , Just $ tshow $ unAccountName $ _kadenaAddress_accountName recipient
+        , case _kadenaAddress_accountCreated recipient of
+          AccountCreated_Yes -> Nothing
```
mankyKitty: using `mempty` or `Text.empty` rather than having to go via `catMaybes` ?

frontend/src/Frontend/Wallet.hs
```diff
@@ -84,9 +84,9 @@ import Frontend.Network
 accountToKadenaAddress :: Account key -> KadenaAddress
 accountToKadenaAddress a = mkKadenaAddress isCreated (_account_chainId a) (_account_name a)
   where
-    isCreated = if unAccountName (_account_name a) == keyToText (_keyPair_publicKey $ _account_key a)
-      then AccountCreated_No -- Wallet only account
-      else AccountCreated_Yes -- Vanity account
+    isCreated = case _account_balance a of
+      Just _ -> AccountCreated_Yes
+      Nothing -> AccountCreated_No
```
mankyKitty: `maybe` function over manual `case` expression?

```
isCreated = maybe AccountCreated_No (const AccountCreated_Yes) $ _account_balance a
```

frontend/src/Frontend/UI/Dialogs/Send.hs
tomsmalley: Nice

PR #264, opened 2019-11-25
[WIP!!!!] feat: First draft of being able to sign with non bip32 keys

benkolera:
I spent some time experimenting with what it would mean to have non wallet keys in chainweaver. This is where my thoughts are up to. 

High level thoughts: 
- It doesn't look *super* ugly and appears that it should work. There's just a bit of a dance keeping cryptonite stuff out of the ghcjs frontend
- We probably don't need to take the public key as the elliptic curve functions can derive the pub key from the private ( https://github.com/kadena-io/pact/blob/69de6d7f92db1321c56c6a06a7040c25c44934a7/src-ghc/Pact/Types/Crypto.hs#L363 )
- Can't fit this into the cardano wallet so we're rolling new PBKDF2 encryption of the secret. We will also have to replicate the cardano wallet's decryption into special short lived memory? 
- Obviously still some work to do!!

### DIFF
--- a/frontend/src/Frontend/UI/Dialogs/ImportLegacyAccount.hs
+++ b/frontend/src/Frontend/UI/Dialogs/ImportLegacyAccount.hs
@@ -90,6 +90,8 @@ uiImportLegacyAccountSettings ideL mChainId initialNotes = Workflow $ do
 
   let eNewAccountOk = fmapMaybe hush $ current dCreateInput <@ eOnImport
 
+  performEvent $ liftIO . print <$> eNewAccountOk
+
   pure
     ( ( "Import Legacy Account"
       , (mempty & walletCfg_importPactKeypair .~ eNewAccountOk
--- a/frontend/src/Frontend/UI/Dialogs/ImportLegacyAccount.hs
+++ b/frontend/src/Frontend/UI/Dialogs/ImportLegacyAccount.hs
@@ -1,0 +1,63 @@
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE RecursiveDo #-}
+{-# LANGUAGE TupleSections #-}
+module Frontend.UI.Dialogs.ImportLegacyAccount
+  ( uiImportLegacyAccountSettings
+  ) where
+
+import Data.Text (Text)
+import Reflex
+import Reflex.Dom.Contrib.CssClass (renderClass)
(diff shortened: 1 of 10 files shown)


frontend/src/Frontend/Wallet.hs
```diff
@@ -190,20 +193,27 @@ makeWallet model conf = do
...
     keys <- foldDyn id initialKeys $ leftmost
       [ ffor onNewKey $ snocIntMap . SomeAccount_Account
       , ffor (_walletCfg_importAccount conf) $ snocIntMap . SomeAccount_Account
       , ffor onDelKey $ \i -> IntMap.insert i SomeAccount_Deleted
       , ffor onWOAccountCreate $ snocIntMap . SomeAccount_Account
+      -- TODO: This eats up indexes for non wallet keys, which could lead to some bad UX
+      -- making it even harder to recover a wallet. May be worth separating them in local
+      -- storage and zipping them together somehow. Does anything out of here rely on the
+      -- stability of the indexes?
+      , ffor onNewImportEd25519 $ snocIntMap . SomeAccount_Account
```
mankyKitty: - I agree that we should store these separately, especially if we're treating them as a legacy thing. 
- I don't know of anything that depends on the stability of the keys.

frontend/src/Frontend/UI/Dialogs/ImportLegacyAccount.hs
```diff
@@ -0,0 +1,133 @@
...
+
+  let isDisabled = either (const True) (const False) <$> dCreateInput
+
+  eNewAccount <- modalFooter $ uiButtonDyn
+    (def & uiButtonCfg_class .~ "button_type_confirm" & uiButtonCfg_disabled .~ isDisabled)
+    (text "Import")
+
+  let eNewAccountOk = fmapMaybe hush $ current dCreateInput <@ eNewAccount
+
+  performEvent $ (liftIO $ putStrLn "BUTTS") <$ eNewAccountOk
```
tomsmalley: :joy:

benkolera: As @mankyKitty can attest, this is a standard, "ben was here" breadcrumb. Lol. Gone now.

benkolera:
Closing this. We don't need to store non-bip32 keys for now.

PR #271, opened 2019-11-26
Blk@redesign/capabilities ui

benkolera:
Done:
- Hides the empty capability line behind the addButton
- Signers list coming from capability and also the checkbox UI
- Apply All button to apply gas capability account to all other capabilities
- Drop restriction of only showing accounts for the selected chain in signing ui
- Puts the signing warp server in dev desktop setup
- Cuts fullDeploy out of signing dialog as uiDeploymentSettings is enough now
- Trims fullDeploy down (don't need withSubmit anymore with signing self sufficient)

Todo:
- Full testing (I've played parrots a bit, but want to do deployments and poke around a lot)

Later (I'll make an Asana for it):
- The signing server seems to get stuck when ob restarts

PR #272, opened 2019-11-26
feat: Refresh button for wallet. Only refresh current network

benkolera:


desktop/src/Desktop/Frontend.hs
```diff
@@ -161,8 +161,8 @@ bipWallet appCfg = do
 
 -- | Returns an event which fires at the given check interval when the user has
 -- been inactive for at least the given timeout.
-watchInactivity :: MonadWidget t m => NominalDiffTime -> NominalDiffTime -> m (Event t ())
-watchInactivity checkInterval timeout = do
+_watchInactivity :: MonadWidget t m => NominalDiffTime -> NominalDiffTime -> m (Event t ())
```
benkolera, line 164: Just getting rid of the warning.

frontend/src/Frontend/UI/Wallet.hs
```diff
@@ -144,9 +156,11 @@ uiKeyItems model = do
 
     el "tbody" $ do
       events <- listWithKey keyMap (uiKeyItem model)
-      dyn_ $ ffor keyMap $ \keys -> when (Map.null keys) $
-        elClass "tr" "wallet__table-row" $ elAttr "td" ("colspan" =: "6" <> "class" =: "wallet__table-cell") $
-          text "No accounts ..."
+      let selectedNetwork = model ^. network_selectedNetwork
+      dyn_ $ ffor2 keyMap selectedNetwork $ \keys net ->
+        unless (any (isJust . activeAccountOnNetwork net) $ Map.elems keys) $
```
benkolera, line 161: Prior to this, if you had no account on the current network but had other accounts (or deleted accounts on this network) you'd have a blank table but with no "no accounts" message.

frontend/src/Frontend/Wallet.hs
```diff
@@ -125,6 +128,11 @@ instance FromJSON key => FromJSON (SomeAccount key) where
   parseJSON Null = pure SomeAccount_Deleted
   parseJSON x = SomeAccount_Account <$> parseJSON x
 
+activeAccountOnNetwork :: NetworkName -> SomeAccount key -> Maybe (Account key)
+activeAccountOnNetwork net = someAccount
```
benkolera, line 132: Deployment settings should use this once #271 is merged.

tomsmalley: Does this need taking care of in this PR?

benkolera: Not really. It shouldn't hold this up. It was just a nice to do given it abstracts it.

frontend/src/Frontend/UI/Wallet.hs
alexfmpe: Shouldn't we feed a filtered version of the map to both the list widget and the message widget instead of doing the check in two places?

PR #273, opened 2019-11-27
Upgrade Receive Dialog

mankyKitty:
This is a WIP of upgrading the receive dialog from the 'accounts' to be able to receive money from a non-bip32 key that is not stored in the wallet.

~~It's a WIP. so be gentle :hammer_and_wrench:~~

### DIFF
--- a/desktop/src/Desktop/Frontend.hs
+++ b/desktop/src/Desktop/Frontend.hs
@@ -17,6 +17,7 @@ import Control.Lens ((?~))
 import Control.Monad (when, (<=<), guard, void)
 import Control.Monad.Fix (MonadFix)
 import Control.Monad.IO.Class
+import GHC.Generics (Generic)
 import Data.Bitraversable
 import Data.Bits ((.|.))
 import Data.Bool (bool)
@@ -31,6 +32,7 @@ import qualified Cardano.Crypto.Wallet as Crypto
 import qualified Control.Newtype.Generics as Newtype
 import qualified Data.Aeson as Aeson
 import qualified Data.ByteString as BS
+import qualified Data.ByteString.Base16 as B16
 import qualified Data.ByteString.Lazy as LBS
 import qualified Data.Text.Encoding as T
 import qualified Data.Text.Encoding.Error as T
@@ -40,6 +42,11 @@ import qualified GHCJS.DOM.EventM as EventM
 import qualified GHCJS.DOM.GlobalEventHandlers as GlobalEventHandlers
 import qualified System.Directory as Directory
 import qualified System.FilePath as FilePath
+
+import qualified Pact.Types.Crypto as PactCrypto
+import Pact.Types.Scheme (PPKScheme)
+import Pact.Types.Util (parseB16TextOnly)
+import qualified Pact.Types.Hash as Pact
 
 import Common.Api (getConfigRoute)
 import Common.Route
@@ -92,15 +99,62 @@ fileStorage dir = Storage
     where path :: Show a => a -> FilePath
           path k = dir </> FilePath.makeValid (show k)
 
-bipCrypto :: Crypto.XPrv -> Text -> Crypto Crypto.XPrv
+newtype XPactSecret = XPactSecret { unXPactSecret :: BS.ByteString }
+
+instance Aeson.ToJSON XPactSecret where
+  toJSON = Aeson.toJSON . T.decodeUtf8 . B16.encode . unXPactSecret
+
+instance Aeson.FromJSON XPactSecret where
+  -- TODO: Should there be some verification of length like XPrv?
+  parseJSON = fmap (XPactSecret . fst . B16.decode . T.encodeUtf8) . Aeson.parseJSON
+
+data DesktopKey
+  = DesktopKeyBip32 Crypto.XPrv
+  | DesktopKeyPact PPKScheme PublicKey XPactSecret
+  deriving (Generic)
+
+-- TODO: This is not backwards compatible. Do we care?
+instance Aeson.FromJSON DesktopKey
+instance Aeson.ToJSON DesktopKey
+
+-- The pact 3.0 blog said that pact supports ethereum ECDSA and ED25519 keys. We probably need
+-- to support both here. See https://medium.com/kadena-io/announcing-pact-3-0-4b0a8f35e6a0
+bipCrypto :: Crypto.XPrv -> Text -> Crypto DesktopKey
 bipCrypto root pass = Crypto
-  { _crypto_sign = \bs k -> pure $ Newtype.pack $ Crypto.unXSignature $ Crypto.sign (T.encodeUtf8 pass) k bs
-  , _crypto_genKey = \i -> do
-    liftIO $ putStrLn $ "Deriving key at index: " <> show i
-    let xprv = Crypto.deriveXPrv scheme (T.encodeUtf8 pass) root (mkHardened $ fromIntegral i)
-    pure (xprv, unsafePublicKey $ Crypto.xpubPublicKey $ Crypto.toXPub xprv)
+  { _crypto_sign = \bs -> \case
+    DesktopKeyBip32 xprv ->
+      pure $ Newtype.pack $ Crypto.unXSignature $ Crypto.sign (T.encodeUtf8 pass) xprv bs
+    DesktopKeyPact pkScheme pubKey encryptedSecret -> do
+      -- TODO TODO TODO : Still no PBKDF2 decryption from the secret key yet.
+      let someKpE = importKey pkScheme (Just $ Newtype.unpack pubKey) (unXPactSecret encryptedSecret)
+      case someKpE of
+        -- TODO: Hash is supposed to be a Base64 encoded bytestring. This likely isn't right :)
+        Right someKp -> liftIO $ Newtype.pack <$> PactCrypto.sign someKp (Pact.Hash bs)
+        Left e -> error $ "Error importing pact key from account: " <> e
+  , _crypto_genKey = \case
+    GenWalletIndex i -> do
+      liftIO $ putStrLn $ "Deriving key at index: " <> show i
+      let xprv = Crypto.deriveXPrv scheme (T.encodeUtf8 pass) root (mkHardened $ fromIntegral i)
+      pure (DesktopKeyBip32 xprv, unsafePublicKey $ Crypto.xpubPublicKey $ Crypto.toXPub xprv)
+    -- This is a little bastardised now as we aren't really generating anything: we are just
+    -- encrypting the secret for serialisation.
+    GenFromPactKey pactKey -> do
+      let
+        pubKey = _pactKey_publicKey pactKey
+        encryptedSecret = XPactSecret (_pactKey_secret pactKey) --TODO TODO TODO Not encrypted yet!!
+      pure (DesktopKeyPact (_pactKey_scheme pactKey) pubKey encryptedSecret, pubKey)
+  -- This assumes that the secret is already base16 encoded (being pasted in, so makes sense)
+  , _crypto_verifyPactKey = \pkScheme sec -> pure $ do
+      secBytes <- parseB16TextOnly sec
+      somePactKey <- importKey pkScheme Nothing secBytes
+      pure $ PactKey pkScheme (unsafePublicKey $ PactCrypto.getPublic somePactKey) secBytes
   }
   where
+    importKey pkScheme mPubBytes secBytes = PactCrypto.importKeyPair
+      (PactCrypto.toScheme pkScheme)
+      (PactCrypto.PubBS <$> mPubBytes)
+      (PactCrypto.PrivBS secBytes)
+
     scheme = Crypto.DerivationScheme2
     mkHardened = (0x80000000 .|.)
 
@@ -140,7 +194,7 @@ bipWallet
      , HasConfigs m
      , HasStorage m, HasStorage (Performable m)
      )
-  => AppCfg Crypto.XPrv t (RoutedT t (R FrontendRoute) (CryptoT Crypto.XPrv m))
+  => AppCfg DesktopKey t (RoutedT t (R FrontendRoute) (CryptoT DesktopKey m))
   -> RoutedT t (R FrontendRoute) m ()
 bipWallet appCfg = do
   mRoot <- getItemStorage localStorage BIPStorage_RootKey
--- a/desktop/src/Desktop/Frontend.hs
+++ b/desktop/src/Desktop/Frontend.hs
@@ -17,7 +17,6 @@ import Control.Lens ((?~))
 import Control.Monad (when, (<=<), guard, void)
 import Control.Monad.Fix (MonadFix)
 import Control.Monad.IO.Class
-import GHC.Generics (Generic)
 import Data.Bitraversable
 import Data.Bits ((.|.))
 import Data.Bool (bool)
@@ -32,7 +31,6 @@ import qualified Cardano.Crypto.Wallet as Crypto
 import qualified Control.Newtype.Generics as Newtype
 import qualified Data.Aeson as Aeson
 import qualified Data.ByteString as BS
-import qualified Data.ByteString.Base16 as B16
 import qualified Data.ByteString.Lazy as LBS
 import qualified Data.Text.Encoding as T
 import qualified Data.Text.Encoding.Error as T
@@ -44,7 +42,6 @@ import qualified System.Directory as Directory
 import qualified System.FilePath as FilePath
 
 import qualified Pact.Types.Crypto as PactCrypto
-import Pact.Types.Scheme (PPKScheme)
 import Pact.Types.Util (parseB16TextOnly)
 import qualified Pact.Types.Hash as Pact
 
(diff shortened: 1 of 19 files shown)


frontend/src/Frontend/Wallet.hs
```diff
@@ -308,6 +321,33 @@ storeKeys ks = setItemStorage localStorage StoreWallet_Keys ks
 loadKeys :: (FromJSON key, HasStorage m, MonadJSM m) => m (Maybe (Accounts key))
 loadKeys = getItemStorage localStorage StoreWallet_Keys
 
+uiAccountNameInput
```
benkolera: This appears to be a UI thing in Frontend.Wallet Probs should go in Frontend.UI.*

frontend/src/Frontend.hs
```diff
@@ -53,7 +54,7 @@ frontend = Frontend
   , _frontend_body = prerender_ loaderMarkup $ do
     (fileOpened, triggerOpen) <- openFileDialog
     let store = browserStorage
-        crypto = Crypto mkSignature (const genKeyPair)
+        crypto = Crypto mkSignature (const genKeyPair) deriveKeyPairFromPrivateKey
```
benkolera: Don't forget the same change in Desktop bipWallet! You are then gonna have to deal with XPrv there so it'll be harder. 

It may actually be worth just using the Pact / Cryptonite stuff there given we had to change the crypto api anyway...

https://github.com/kadena-io/chainweaver/pull/264/files#diff-152ee524a8f6733eb8014b53f0ebbf73R150

frontend/src/Frontend/UI/Widgets.hs
```diff
@@ -372,6 +374,30 @@ noAutofillAttrs = Map.fromList
...
+     , MonadHold t m
+     , MonadFix m
+     , PostBuild t m
+     )
+  => CssClass
+  -> (Dynamic t (Text -> Either Text a))
+  -- ^ Validation function returning `Just error message` on error.
+  -> m (InputElement EventResult (DomBuilderSpace m) t)
+  -> m (Event t a)
+validatedInput uCls check input = divClass (renderClass uCls) $ do
```
benkolera: You made this and didn't end up using it?

frontend/src/Frontend/UI/Dialogs/Receive.hs
```diff
@@ -1,30 +1,254 @@
-
+{-# LANGUAGE RecursiveDo #-}
```
benkolera, line 1: I'm a bit suspicious of why we can't use uiDeploymentSettings here like the signing ui does. But I trust you and we're out of time to dig deeply. Code dupe and cleanup later if possible sounds fine. :slightly_smiling_face:

mankyKitty: This is a suspicion that I share. The original UX flow did not allow for the deployment process to be used. That still holds but I'm wary of _not_ using `uiDeploymentSettings` for obvious reasons. Definitely something for me to follow up on.

frontend/src/Frontend/UI/Dialogs/Receive.hs
```diff
@@ -1,62 +1,445 @@
...
+      [ text <$> deriveErr
+      , blank <$ okPair
+      ]
+
+    keyPair <- holdDyn Nothing $ Just <$> okPair
+
+    chain <- divClass "account-details__receive-from-chain" $ userChainIdSelect model
+
+    amount <- fst . snd <$> mkLabeledInput True "Amount"
+      (uiRealWithPrecisionInputElement maxCoinPrecision id) def
```
alexfmpe: #278 renames this function (semantic conflict)

frontend/src/Frontend/UI/Dialogs/Receive.hs
```diff
@@ -1,62 +1,445 @@
...
+    displayText lbl v cls =
+      let
+        attrFn cfg = uiInputElement $ cfg
+          & initialAttributes <>~ ("disabled" =: "true" <> "class" =: (" " <> cls))
+      in
+        mkLabeledInputView True lbl attrFn $ pure v
+
+  (showingAddr, (conf, tfrInfo)) <- divClass "modal__main account-details" $ do
+    rec
+      showingKadenaAddress <- holdDyn True $
```
alexfmpe: `showingKadenaAddress <- toggle True $ onAddrClick <> onReceiClick`

frontend/src/Frontend/Wallet.hs
```diff
@@ -81,10 +86,18 @@ import Frontend.KadenaAddress
 import Frontend.Storage
 import Frontend.Network
 
+import Frontend.UI.Widgets
+
+accountIsCreated :: Account key -> AccountCreated
+accountIsCreated = maybe AccountCreated_No (const AccountCreated_Yes) . _account_balance
+
+accountCreatedYesNo :: Account key -> a -> a -> a
+accountCreatedYesNo a yes no = case accountIsCreated a of
```
alexfmpe: How about `accountCreatedBool :: a -> a -> Account key -> a` (and flip the yes/no order) to match `bool` ?

frontend/src/Frontend/Wallet.hs
```diff
@@ -81,10 +86,18 @@ import Frontend.KadenaAddress
 import Frontend.Storage
 import Frontend.Network
 
+import Frontend.UI.Widgets
+
+accountIsCreated :: Account key -> AccountCreated
+accountIsCreated = maybe AccountCreated_No (const AccountCreated_Yes) . _account_balance
```
alexfmpe, line 90: Any reason not to use `newtype AccountCreated = AccountCreated Bool` instead?
Then you could simply `accountIsCreated = AccountCreated . isJust . _account_balance`

frontend/src/Frontend/UI/Dialogs/Receive.hs
```diff
@@ -1,62 +1,446 @@
...
+
+    infos = (liftA2 . liftA2) (,) netInfo transferInfo
+
+  pure ( (conf, onClose <> done)
+       , uncurry
+         <$> current (receiveFromLegacySubmit account chain <$> ttl <*> gaslimit)
+         <@> tagMaybe (current infos) deploy
+       )
+
+-- uiReceiveDeployLegacyTransfer
```
alexfmpe: What's going on here?

mankyKitty: Nothing to see here, move along ...

frontend/src/Frontend/UI/Dialogs/Receive.hs
mankyKitty: nice, ty

desktop/src/Desktop/Frontend.hs
```diff
@@ -92,15 +99,62 @@ fileStorage dir = Storage
...
+  -- TODO: Should there be some verification of length like XPrv?
+  parseJSON = fmap (XPactSecret . fst . B16.decode . T.encodeUtf8) . Aeson.parseJSON
+
+data DesktopKey
+  = DesktopKeyBip32 Crypto.XPrv
+  | DesktopKeyPact PPKScheme PublicKey XPactSecret
+  deriving (Generic)
+
+-- TODO: This is not backwards compatible. Do we care?
+instance Aeson.FromJSON DesktopKey
```
benkolera: We want a backwards compat non derived instance here esp because we never want to store these new keys.

desktop/src/Desktop/Frontend.hs
```diff
@@ -92,15 +99,62 @@ fileStorage dir = Storage
...
+instance Aeson.ToJSON DesktopKey
+
+-- The pact 3.0 blog said that pact supports ethereum ECDSA and ED25519 keys. We probably need
```
benkolera: We don't need this comment. Doug confirmed ED25519 keys only.

desktop/src/Desktop/Frontend.hs
```diff
@@ -92,15 +99,62 @@ fileStorage dir = Storage
...
-  { _crypto_sign = \bs k -> pure $ Newtype.pack $ Crypto.unXSignature $ Crypto.sign (T.encodeUtf8 pass) k bs
-  , _crypto_genKey = \i -> do
-    liftIO $ putStrLn $ "Deriving key at index: " <> show i
-    let xprv = Crypto.deriveXPrv scheme (T.encodeUtf8 pass) root (mkHardened $ fromIntegral i)
-    pure (xprv, unsafePublicKey $ Crypto.xpubPublicKey $ Crypto.toXPub xprv)
+  { _crypto_sign = \bs -> \case
+    DesktopKeyBip32 xprv ->
+      pure $ Newtype.pack $ Crypto.unXSignature $ Crypto.sign (T.encodeUtf8 pass) xprv bs
+    DesktopKeyPact pkScheme pubKey encryptedSecret -> do
+      -- TODO TODO TODO : Still no PBKDF2 decryption from the secret key yet.
```
benkolera: Need to tidy up the names and comments here because this thing isn't and isn't planned to be encrypted.

desktop/src/Desktop/Frontend.hs
```diff
@@ -92,15 +99,62 @@ fileStorage dir = Storage
...
+        Right someKp -> liftIO $ Newtype.pack <$> PactCrypto.sign someKp (Pact.Hash bs)
+        Left e -> error $ "Error importing pact key from account: " <> e
+  , _crypto_genKey = \case
+    GenWalletIndex i -> do
+      liftIO $ putStrLn $ "Deriving key at index: " <> show i
+      let xprv = Crypto.deriveXPrv scheme (T.encodeUtf8 pass) root (mkHardened $ fromIntegral i)
+      pure (DesktopKeyBip32 xprv, unsafePublicKey $ Crypto.xpubPublicKey $ Crypto.toXPub xprv)
+    -- This is a little bastardised now as we aren't really generating anything: we are just
+    -- encrypting the secret for serialisation.
+    GenFromPactKey pactKey -> do
```
benkolera: This branch should error. We absolutely never ever want to store these keys.

desktop/src/Desktop/Frontend.hs
```diff
@@ -92,15 +99,62 @@ fileStorage dir = Storage
...
+  | DesktopKeyPact PPKScheme PublicKey XPactSecret
+  deriving (Generic)
+
+-- TODO: This is not backwards compatible. Do we care?
+instance Aeson.FromJSON DesktopKey
+instance Aeson.ToJSON DesktopKey
+
+-- The pact 3.0 blog said that pact supports ethereum ECDSA and ED25519 keys. We probably need
+-- to support both here. See https://medium.com/kadena-io/announcing-pact-3-0-4b0a8f35e6a0
+bipCrypto :: Crypto.XPrv -> Text -> Crypto DesktopKey
```
benkolera: Not for now, but we should drop in an asana ticket to maybe separate the type for storage vs signing. @tomsmalley  suggested maybe doing an associated type in the crypto class and that sounds pretty good to me.

frontend/src/Frontend.hs
```diff
@@ -53,7 +53,7 @@ frontend = Frontend
   , _frontend_body = prerender_ loaderMarkup $ do
     (fileOpened, triggerOpen) <- openFileDialog
     let store = browserStorage
-        crypto = Crypto mkSignature (const genKeyPair)
+        crypto = Crypto mkSignature (const genKeyPair) (\_ _ -> pure $ Left "not supported")
```
benkolera: You could use Frontend.Crypto.Ed25519.deriveKeyPairFromPrivateKey here, but it probably doesn't matter since it wont get used anyway.

desktop/src/Desktop/Frontend.hs
```diff
@@ -92,15 +99,62 @@ fileStorage dir = Storage
...
-    pure (xprv, unsafePublicKey $ Crypto.xpubPublicKey $ Crypto.toXPub xprv)
+  { _crypto_sign = \bs -> \case
+    DesktopKeyBip32 xprv ->
+      pure $ Newtype.pack $ Crypto.unXSignature $ Crypto.sign (T.encodeUtf8 pass) xprv bs
+    DesktopKeyPact pkScheme pubKey encryptedSecret -> do
+      -- TODO TODO TODO : Still no PBKDF2 decryption from the secret key yet.
+      let someKpE = importKey pkScheme (Just $ Newtype.unpack pubKey) (unXPactSecret encryptedSecret)
+      case someKpE of
+        -- TODO: Hash is supposed to be a Base64 encoded bytestring. This likely isn't right :)
+        Right someKp -> liftIO $ Newtype.pack <$> PactCrypto.sign someKp (Pact.Hash bs)
```
benkolera: I'm surprised and terrified that that actually worked. Lol. Probably remove the TODO though if it worked. :)

benkolera (changes requested):

desktop/src/Desktop/Frontend.hs
```diff
@@ -92,15 +99,62 @@ fileStorage dir = Storage
     where path :: Show a => a -> FilePath
           path k = dir </> FilePath.makeValid (show k)
 
-bipCrypto :: Crypto.XPrv -> Text -> Crypto Crypto.XPrv
+newtype XPactSecret = XPactSecret { unXPactSecret :: BS.ByteString }
```
tomsmalley: what is this, is it different from `PrivateKey`?

desktop/src/Desktop/Frontend.hs
```diff
@@ -92,15 +99,62 @@ fileStorage dir = Storage
...
+
+instance Aeson.ToJSON XPactSecret where
+  toJSON = Aeson.toJSON . T.decodeUtf8 . B16.encode . unXPactSecret
+
+instance Aeson.FromJSON XPactSecret where
+  -- TODO: Should there be some verification of length like XPrv?
+  parseJSON = fmap (XPactSecret . fst . B16.decode . T.encodeUtf8) . Aeson.parseJSON
+
+data DesktopKey
```
tomsmalley: The presence of this seems odd if we aren't storing things. Like you said in mattermost, signing with these keys can be done by changing `buildCmd` only.

desktop/src/Desktop/Frontend.hs
tomsmalley: Having to error out in these branches definitely means we shouldn't go this route

frontend/src/Frontend/Crypto/Class.hs
```diff
@@ -15,19 +15,44 @@ import Obelisk.Route.Frontend
...
+
+data GenKeyArg
+  = GenWalletIndex Int
+  | GenFromPactKey PactKey
 
 data Crypto key = Crypto
   { _crypto_sign :: ByteString -> key -> JSM Signature
-  , _crypto_genKey :: Int -> JSM (key, PublicKey)
+  , _crypto_genKey :: GenKeyArg -> JSM (key, PublicKey)
+  , _crypto_verifyPactKey :: PPKScheme -> Text -> JSM (Either String PactKey)
```
tomsmalley, line 37: This seems like it doesn't belong here. Does this function change between desktop/web?

frontend/src/Frontend/Crypto/Class.hs
```diff
@@ -15,19 +15,44 @@ import Obelisk.Route.Frontend
...
+  } deriving Show
+
+data GenKeyArg
+  = GenWalletIndex Int
+  | GenFromPactKey PactKey
 
 data Crypto key = Crypto
   { _crypto_sign :: ByteString -> key -> JSM Signature
-  , _crypto_genKey :: Int -> JSM (key, PublicKey)
+  , _crypto_genKey :: GenKeyArg -> JSM (key, PublicKey)
```
tomsmalley: Shouldn't need to do this if we change `buildCmd`

frontend/src/Frontend/KadenaAddress.hs
```diff
@@ -107,11 +109,12 @@ data KadenaAddressPrefix
...
   deriving (Show, Eq)
 
-data AccountCreated
-  = AccountCreated_Yes
-  | AccountCreated_No
+newtype AccountCreated = AccountCreated Bool
   deriving (Show, Eq)
 
+accountCreatedBool :: a -> a -> AccountCreated -> a
+accountCreatedBool no yes (AccountCreated b) = bool no yes b
```
tomsmalley: Boolean blindness here is terrible, I'd rather just pattern match anywhere this is used

frontend/src/Frontend/UI/Dialogs/Receive.hs
```diff
@@ -1,62 +1,337 @@
...
+        }
+      , _scArgs =
+        [ PLiteral $ LString $ unAccountName sender
+        , PLiteral $ LString $ unAccountName $ _account_name account
+        , PLiteral $ LDecimal amount
+        ]
+      }
+
+    transferCapability = DappCap
+      { _dappCap_role = "TRANFER"
```
tomsmalley: Spelling

frontend/src/Frontend/UI/Dialogs/Receive.hs
```diff
@@ -1,62 +1,337 @@
...
+      { _pmChainId = chain
+      , _pmSender = unAccountName sender
+      , _pmGasLimit = gasLimit
+      , _pmTTL = ttl
+      }
 
-  done <- modalFooter $ confirmButton def "Done"
-  pure (mempty, close <> done)
+  cmd <- buildCmd Nothing networkId pm signingPairs [] code dat pkCaps
+  pure $ ReceiveBuildCommandInfo code transferInfo cmd signingPairs pm caps pkCaps dat
```
tomsmalley: Why package all this info up and then only use the command at the use site?

tomsmalley (changes requested):

frontend/src/Frontend/Crypto/Class.hs
tomsmalley: Yes, it does change: on desktop we use the function from Pact.Types.Crypto which is only available on GHC.

desktop/src/Desktop/Frontend.hs
mankyKitty: It's going away, along with `DesktopKey`

desktop/src/Desktop/Frontend.hs
mankyKitty: Removing

frontend/src/Frontend/Crypto/Class.hs
mankyKitty: Reverting

desktop/src/Desktop/Frontend.hs
mankyKitty: Ticket created

benkolera (approved):

PR #278, opened 2019-11-27
Clamp values in transaction settings fields

alexfmpe:


frontend/src/Frontend/UI/Widgets.hs
```diff
@@ -176,59 +177,102 @@ uiRealInputElement cfg = do
     inputElement $ cfg & initialAttributes %~
         (<> ("type" =: "number")) . addInputElementCls . addNoAutofillAttrs
 
+uiCorrectingInputElement
```
alexfmpe, line 180: > I'll just factor out the stuff from `uiRealWithPrecisionInputElement`, how hard can it be

frontend/src/Frontend/UI/Widgets.hs
```diff
@@ -176,59 +177,102 @@ uiRealInputElement cfg = do
...
-uiRealWithPrecisionInputElement
+uiNonnegativeRealWithPrecisionInputElement
   :: forall t m er a. (DomBuilder t m, MonadFix m, MonadHold t m)
   => Word8
   -> (Decimal -> a)
   -> InputElementConfig er t (DomBuilderSpace m)
-  -> m (InputElement er (DomBuilderSpace m) t, (Dynamic t (Maybe a), Event t a))
-uiRealWithPrecisionInputElement prec fromDecimal cfg = do
+  -> m (InputElement er (DomBuilderSpace m) t, Dynamic t (Maybe a), Event t a)
+uiNonnegativeRealWithPrecisionInputElement prec fromDecimal cfg = do
```
alexfmpe, line 222: I hate this name

frontend/src/Frontend/UI/Widgets.hs
```diff
@@ -176,59 +177,102 @@ uiRealInputElement cfg = do
...
+
+    sanitize :: Decimal -> Maybe (Decimal, Text)
+    sanitize decimal = asum
+      [ (0, "Cannot be negative")
+        <$ guard (decimal < 0)
+      , (D.roundTo prec decimal, ("Rounded to " <> tshow prec <> " places"))
+        <$ guard (D.decimalPlaces decimal > prec)
+      ]
 
+-- TODO: correct floating point decimals
```
alexfmpe: `min`/`max` attributes only seem to be applied to arrow-key updates? I could certainly add a minus sign or just add digits to the number without them affecting anything

frontend/src/Frontend/UI/Widgets.hs
```diff
@@ -176,59 +177,102 @@ uiRealInputElement cfg = do
...
+
+    pure (r
+          { _inputElement_value = fmap fixNum $ _inputElement_value r
+          , _inputElement_input = fmap fixNum $ _inputElement_input r
+          }
+         , fst <$> input)
+
   where
     fixNum = T.takeWhile (/='.')
+    numberAttrs = mconcat $ fmapMaybe id
```
alexfmpe, line 265: `min`/`max` attributes only seem to be applied to arrow-key updates? I could certainly add a minus sign or just add digits to the number without the attributes affecting anything

frontend/src/Frontend/UI/Widgets.hs
```diff
@@ -176,59 +177,104 @@ uiRealInputElement cfg = do
...
+  ie <- inputElement $ cfg & inputElementConfig_setValue %~ (\e -> leftmost [attemptCorrection e, forceCorrection inp])
+  let
+    inp = _inputElement_input ie
+    val = value ie
+
+    sanitization :: Functor f => f Text -> f (Maybe (a, Maybe explanation))
+    sanitization = fmap $ \i -> ffor (parse i) $ \p ->
+      case sanitize p of
+        Nothing -> (p, Nothing)
+        Just (a, x) -> (a, Just x)
```
mankyKitty, line 198: Depending on your taste for Functor for tuples. ;)
```
maybe (p, Nothing) (fmap Just) $ sanitize p
```

frontend/src/Frontend/UI/Widgets.hs
benkolera: Yeah, I think so. That's what I'd expect. It's not really validation.

benkolera (approved):

frontend/src/Frontend/UI/Widgets.hs
tomsmalley: It gets slightly better if you change nonnegative -> positive, but yeah

frontend/src/Frontend/UI/Widgets.hs
```diff
@@ -176,59 +177,104 @@ uiRealInputElement cfg = do
...
+uiNonnegativeRealWithPrecisionInputElement
   :: forall t m er a. (DomBuilder t m, MonadFix m, MonadHold t m)
   => Word8
   -> (Decimal -> a)
   -> InputElementConfig er t (DomBuilderSpace m)
-  -> m (InputElement er (DomBuilderSpace m) t, (Dynamic t (Maybe a), Event t a))
-uiRealWithPrecisionInputElement prec fromDecimal cfg = do
+  -> m (InputElement er (DomBuilderSpace m) t, Dynamic t (Maybe a), Event t a)
+uiNonnegativeRealWithPrecisionInputElement prec fromDecimal cfg = do
   rec
```
tomsmalley, line 223: Seems like the `rec` is no longer needed

tomsmalley (commented):
Just some nitpicks

frontend/src/Frontend/UI/Widgets.hs
alexfmpe: I was tempted to do this, but `second` instead of `fmap`

frontend/src/Frontend/UI/Widgets.hs
alexfmpe: Ah good catch

frontend/src/Frontend/UI/Widgets.hs
alexfmpe: It accepts 0 though.

PR #279, opened 2019-11-27
Fix styling of 'Amount' field in send dialog

alexfmpe:
On top of #278 because conflicts

### DIFF
--- a/frontend/src/Frontend/Network.hs
+++ b/frontend/src/Frontend/Network.hs
@@ -7,6 +7,7 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NumDecimals #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE PatternGuards #-}
 {-# LANGUAGE QuasiQuotes #-}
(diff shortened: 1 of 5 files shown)


frontend/src/Frontend/Network.hs
```diff
@@ -394,9 +396,12 @@ maxCoinPrecision = 12
 defaultTransactionTTL :: TTLSeconds
 defaultTransactionTTL = TTLSeconds (8 * 60 * 60) -- 8 hours
 
--- | TODO: Better defaults!!!
+-- Taken from https://github.com/kadena-io/chainweb-node/blob/85688ea0182d1b1ab0d8d784a48b4851a950ec7a/src/Chainweb/Chainweb.hs#L344
+chainwebGasLimit :: Num a => a
+chainwebGasLimit = 1e5
```
benkolera, line 401: Is the default that I set really the limit? Ouch.

benkolera (approved):

frontend/src/Frontend/Network.hs
tomsmalley: Ha, yeah, maybe we should drop the default a little? Unless that requirement came from kadena

alexfmpe: The limit will change to 1e6 soon (https://github.com/kadena-io/chainweb-node/commit/cf83c0f393825a1958e4c1a4007f90606f867e78#diff-1f9807c50e8c60a1c91ca0f4708cb6d1R347).

PR #280, opened 2019-11-28
Crosschain transfers

tomsmalley:
The branch this is based on is a little behind.

Need to test:
- transfer-crosschain to accounts not in wallet which do/don't exist

Need to do:
- ~~rebase~~
- ability to pay gas with Ed25519 keys <- this should be handled in a subsequent PR which uses some of Sean's work
- cleanup error handling
- cleanup UIs

### DIFF
--- a/frontend/src/Frontend/Network.hs
+++ b/frontend/src/Frontend/Network.hs
@@ -52,8 +52,11 @@ module Frontend.Network
   , prettyPrintNetworkErrors
   , networkErrorResultToEither
   , buildCmd
+  , buildContPayload
+  , buildCmdWithPayload
   , mkSimpleReadReq
   , getNetworkNameAndMeta
+  , getCreationTime
     -- * Defaults
   , defaultTransactionGasLimit
   , defaultTransactionGasPrice
@@ -102,7 +105,6 @@ import           Pact.Types.Runtime                (PactError (..), GasLimit (..
 import           Pact.Types.ChainMeta              (PublicMeta (..), TTLSeconds (..), TxCreationTime (..))
 import qualified Pact.Types.ChainMeta              as Pact
 import           Pact.Types.ChainId                (NetworkId (..))
-
 import           Pact.Types.Exp                    (Literal (LString))
 import           Pact.Types.Hash                   (hash, Hash (..), TypedHash (..), toUntypedHash)
 import           Pact.Types.RPC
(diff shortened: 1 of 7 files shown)


frontend/src/Frontend/Network.hs
```diff
@@ -724,7 +726,9 @@ performLocalReadCustom networkL unwrapUsr onReqs = do
   let
     unwrap = map (networkRequest_endpoint .~ Endpoint_Local) . unwrapUsr
     fakeNetwork = networkL & network_meta . mapped . Pact.pmCreationTime .~ time
-  performNetworkRequestCustom fakeNetwork unwrap onReqs
+  resp <- performNetworkRequestCustom fakeNetwork unwrap onReqs
+  performEvent_ $ ffor resp $ liftIO . print . snd
```
tomsmalley: leftover debugging

frontend/src/Frontend/UI/Wallet.hs
```diff
@@ -188,8 +188,7 @@ uiKeyItem model i d = do
             td $ dynText $ _account_notes <$> account
             td $ dynText $ ffor account $ \a -> case _account_balance a of
               Nothing -> "Unknown"
-              Just b -> tshow (unAccountBalance b) <> " KDA"
-
+              Just b -> tshow (unAccountBalance b) <> " KDA" <> maybe "" (const "*") (_account_unfinishedCrossChainTransfer a)
```
tomsmalley, line 210: Probably should have a tooltip showing what this means. Should run this by Kadena anyway.

benkolera: Yeah, will cover this in the demo.

frontend/src/Frontend/UI/Dialogs/Send.hs
```diff
@@ -203,15 +280,47 @@ sendConfig model sender = Workflow $ do
...
+              else do
+                -- TODO this bit should have an option with a generic input for Ed25519 keys
+                -- and perhaps be skippable entirely in favour of a blob the user
+                -- can send to someone else to continue the tx on the recipient chain
+                elClass "h2" "heading heading_type_h2" $ text $ "Gas Payer (Chain " <> _chainId toChain <> ")"
+                divClass "group" $ elClass "div" "segment segment_type_tertiary labeled-input" $ do
+                  divClass "label labeled-input__label" $ text "Account Name"
+                  let cfg = def & dropdownConfig_attributes .~ pure ("class" =: "labeled-input__input")
+                      chain = pure $ Just toChain
+                  gasAcc <- uiSenderDropdown cfg never model chain
```
benkolera, line 322: This is going to let people choose gas payers that aren't on that chain accidentally, but until we can fix up our model of keys and accounts we've got to roll with this.

benkolera: It'd also be great if we could guarantee that the gas payer had gas money before kicking this off to reduce silly user errors, but not for right now.

benkolera (approved):

frontend/src/Frontend/UI/Dialogs/Send.hs
tomsmalley: Is it? We pass in the chain ID

benkolera: Yup: https://github.com/kadena-io/chainweaver/blob/redesign/frontend/src/Frontend/UI/DeploymentSettings.hs#L715

PR #282, opened 2019-11-28
Handle validation failure errors

alexfmpe:


frontend/src/Frontend/UI/Dialogs/DeployConfirmation.hs
```diff
@@ -281,28 +289,36 @@ submitTransactionWithFeedback cmd chain nodeInfos = do
...
+              -- We should always have a txId when we have a result
+              (Just _txId, Pact.PactResult (Right a)) -> do
+                listen Status_Done
+                setMessage $ Just $ Right a
+                -- TODO wait for confirmation...
+              (_, Pact.PactResult (Left err)) -> do
+                listen Status_Failed
+                setMessage $ Just $ Left $ NetworkError_CommandFailure err
+              -- This case shouldn't happen
+              _ -> listen Status_Failed
```
mankyKitty: Might be worth having something here to report that we've encountered an "impossible" case ?

benkolera (approved):

PR #285, opened 2019-11-29
Make mac app shrinkable

alexfmpe:


mac/cbits/mac.m
```diff
@@ -80,7 +80,6 @@ void resizeWindow(int minWidth, int minHeight) {
   [[NSOperationQueue mainQueue] addOperationWithBlock:^(void) {
     NSApplication *application = [NSApplication sharedApplication];
     NSWindow *window = [application mainWindow];
-    [window setContentMinSize: NSMakeSize(minWidth, minHeight)]; //doesn't seem to change size per-se, only affects resizes
```
benkolera, line 83: Oh, we don't want to constrain things after all?

tomsmalley: This line would stop the user from shrinking the app at all beyond the size limit, so on small screens, a lot of the app was actually off screen (with no way of bringing it on screen).

PR #288, opened 2019-11-29
Fixes to balance retrieval and positive real input

benkolera:
Also on top of #287 because everything is in chaos atm. Lol. 

This disables the filtering of balance refresh to the current network because that was weirdly broken meaning that if you had two networks then one wouldn't get the balances properly even when switching the dropdown. Errors from trying the wrong network are better than missing balances.

Changing the rounding done by #278 to only happen on blur. Much nicer for the user. :)

### DIFF
--- a/frontend/src/Frontend/UI/Widgets.hs
+++ b/frontend/src/Frontend/UI/Widgets.hs
@@ -99,10 +99,28 @@ uiCheckbox
   -> m (Checkbox t)
 uiCheckbox cls b cfg c =
   elKlass "label" (cls <> "label checkbox checkbox_type_secondary") $ do
-    cb <- checkbox b $ cfg
+    cb <- checkbox' b $ cfg
     elClass "span" "checkbox__checkmark checkbox__checkmark_type_secondary" blank
     c
     pure cb
+
+-- | Create an editable checkbox
+--   Note: if the "type" or "checked" attributes are provided as attributes, they will be ignored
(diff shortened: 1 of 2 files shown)


frontend/src/Frontend/UI/Widgets.hs
```diff
@@ -178,50 +198,73 @@ uiRealInputElement cfg = do
         (<> ("type" =: "number")) . addInputElementCls . addNoAutofillAttrs
 
 uiCorrectingInputElement
-  :: forall t m a er explanation. DomBuilder t m
```
benkolera, line 205: Had to make er concrete to get the blur event

benkolera:
@alexfmpe

frontend/src/Frontend/Wallet.hs
```diff
@@ -282,7 +282,7 @@ getBalances model accounts = do
     accountBalanceReq acc = "(coin.get-balance " <> tshow (unAccountName acc) <> ")"
     mkReqs :: (NetworkName, PublicMeta) -> Accounts key -> Performable m (IntMap (SomeAccount key, Maybe NetworkRequest))
     mkReqs meta someaccs = for someaccs $ \sa ->
-      fmap (sa,) . traverse (mkReq meta) $ activeAccountOnNetwork (fst meta) sa
```
benkolera, line 285: This has a bug where it doesn't show balances for the non default wallet. This change reverts the smarts attempted in #272

PR #289, opened 2019-11-29
Signing Screen Fixes

mankyKitty:
Rename 'Signing Accounts' with 'Unrestricted'.
Rename 'Apply all' to 'Apply gas payer to all'.
Hide apply all button unless there is more than one capability.

### DIFF
--- a/frontend/src/Frontend/UI/DeploymentSettings.hs
+++ b/frontend/src/Frontend/UI/DeploymentSettings.hs
@@ -61,7 +61,7 @@ module Frontend.UI.DeploymentSettings
   , Identity (runIdentity)
   ) where
 
-import Control.Applicative ((<|>))
+import Control.Applicative ((<|>),liftA2)
 import Control.Arrow (first, (&&&))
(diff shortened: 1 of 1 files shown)


frontend/src/Frontend/UI/DeploymentSettings.hs
```diff
@@ -938,17 +942,24 @@ uiSenderCapabilities m cid mCaps mkSender = do
...
-      (text "Apply to all")
-    let eApplyToAll = fmapMaybe id $ current mGasAcct' <@ eClickApplyToAll
+          pure ( _capabilityInputRow_account gas
+               , combineMaps [(_capabilityInputRow_value gas),rest]
+               , constDyn 0
+               )
+
+    -- If the gas capability is set, we enable the button that will set every other
+    -- capability's sender from the gas account.
+    eApplyToAllClick <- (switchHold never =<<) $ dyn $ ffor rowCount $ \n ->
```
alexfmpe, line 952: Could also set `display: none` based on `rowCount` but I guess this works

PR #290, opened 2019-11-29
Align network dropdown with accounts table on wallet screen

mankyKitty:


alexfmpe:
Does this also cause the text inside the dropdown to be vertically aligned with header / table text or is are padding tweaks needed?

mankyKitty:
The text inside the dropdown lines up with the wallet heading on my screen. I don't have a mac to make / test the other changes though

PR #292, opened 2019-12-02
Unify KDA input fields to a single widget, allow for input of '.9' as well as '0.9'

mankyKitty:
Add a single purpose widget to be the go to input field when dealing with the KDA
values. This field contains the 'KDA' prefix on the field, the non-negative precision
handling using the coin precision, and also prevents the use of mouse-wheel or up/down
arrows to manipulate the value.

Call Function amount input field is unaffected.

### DIFF
--- a/frontend/src/Frontend/UI/Dialogs/Receive.hs
+++ b/frontend/src/Frontend/UI/Dialogs/Receive.hs
@@ -110,7 +110,6 @@ uiReceiveFromLegacyAccount model = do
   chain <- divClass "account-details__receive-from-chain" $ userChainIdSelect model
 
   amount <- view _2 <$> mkLabeledInput True "Amount" uiGasPriceInputField def
-    -- (uiNonnegativeRealWithPrecisionInputElement maxCoinPrecision id) def
 
   pure $ (\macc mc mamnt mkeypair -> LegacyTransferInfo <$> macc <*> mc <*> mamnt <*> mkeypair)
     <$> mAccountName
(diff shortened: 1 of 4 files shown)


frontend/src/Frontend/UI/Widgets.hs
```diff
@@ -277,7 +284,12 @@ uiNonnegativeRealWithPrecisionInputElement prec fromDecimal cfg = do
 
   where
     stepSize = "0." <> T.replicate (fromIntegral prec - 1) "0" <> "1"
-    parse = tread
+    -- parse = tread
```
tomsmalley: leftover

tomsmalley (approved):

frontend/src/Frontend/UI/Dialogs/Receive.hs
```diff
@@ -108,8 +109,8 @@ uiReceiveFromLegacyAccount model = do
 
   chain <- divClass "account-details__receive-from-chain" $ userChainIdSelect model
 
-  amount <- view _2 <$> mkLabeledInput True "Amount"
-    (uiNonnegativeRealWithPrecisionInputElement maxCoinPrecision id) def
+  amount <- view _2 <$> mkLabeledInput True "Amount" uiGasPriceInputField def
+    -- (uiNonnegativeRealWithPrecisionInputElement maxCoinPrecision id) def
```
alexfmpe: Leftover

PR #294, opened 2019-12-02
Gate the network activity of the preview pane.

mankyKitty:
The preview pane should only attempt to build a preview when it is the active pane. This
checks the currently selected pane to avoid unnecessary network calls.

### DIFF
--- a/frontend/src/Frontend/UI/DeploymentSettings.hs
+++ b/frontend/src/Frontend/UI/DeploymentSettings.hs
@@ -389,7 +389,6 @@ uiDeploymentSettings m settings = mdo
 
       when (_deploymentSettingsConfig_includePreviewTab settings) $ tabPane mempty curSelection DeploymentSettingsView_Preview $ do
         let currentNode = headMay . rights <$> (m ^. network_selectedNodes)
-
             mNetworkId = (hush . mkNetworkName . nodeVersion =<<) <$> currentNode
 
             mHeadAccount = fmap _account_name . findFirstVanityAccount <$> (m ^. wallet_accounts)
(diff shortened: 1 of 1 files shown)


frontend/src/Frontend/UI/DeploymentSettings.hs
```diff
@@ -398,18 +397,22 @@ uiDeploymentSettings m settings = mdo
...
-          <*> gasLimit
-          <*> ttl
-          <*> code
-          <*> (m ^. network_meta)
-          <*> capabilities
-          <*> (m ^. jsonData . jsonData_data)
-          <*> mNetworkId
-          <*> aChainId
-          <*> aSender
+            preVIEW = uiDeployPreview m settings
```
alexfmpe: What's with the casing?

frontend/src/Frontend/UI/DeploymentSettings.hs
```diff
@@ -1030,7 +1034,7 @@ uiDeployPreview model settings accounts signers gasLimit ttl code lastPublicMeta
       extraSigners = _deploymentSettingsConfig_extraSigners settings
       jsondata = HM.union jsonData0 deploySettingsJsonData
 
-  eCmds <- performEvent $ ffor pb $ \_ -> do
+  eCmds <- performEvent $ ffor (traceEvent "pb" pb) $ \_ -> do
```
alexfmpe: debug leftovers?

frontend/src/Frontend/UI/DeploymentSettings.hs
```diff
@@ -398,18 +397,22 @@ uiDeploymentSettings m settings = mdo
...
+              <*> (m ^. network_meta)
+              <*> capabilities
+              <*> (m ^. jsonData . jsonData_data)
+              <*> mNetworkId
+              <*> aChainId
+              <*> aSender
+
+        dyn_ $ join $ ffor curSelection $ \case
+          DeploymentSettingsView_Preview -> preVIEW
+          _ -> constDyn blank
```
alexfmpe, line 415: Since we're using `dyn_` anyway, no need to provide a dummy payload in the non-preview case:
```haskell
dyn_ $ ffor curSelection $ \case
  DeploymentSettingsView_Preview -> void preVIEW
  _ -> blank
```

PR #295, opened 2019-12-02
Rename 'Address' to 'Kadena Address' on Send dialog

mankyKitty:


alexfmpe:
Does this need to be on top of #294 ?

mankyKitty:
No, it shouldn't. I was a bit too keen. I'll close this and do it properly.

PR #297, opened 2019-12-03
Make more screens in setup workflow accept enter key

alexfmpe:


desktop/src/Desktop/Setup.hs
```diff
@@ -104,18 +104,23 @@ textTo = fromString . T.unpack
 tshow :: Show a => a -> Text
 tshow = T.pack . show
 
--- | Form wrapper which will automatically handle submit on enter so long as you have
--- a submit button in the form. Please track this submit event instead of a button click
--- as it is very easy to put the button outside of the form and break enter handling.
--- This way, both will work or both will be broken (if there is no button in the form)
-form :: forall t m a. DomBuilder t m => Text -> m a -> m (Event t (), a)
-form c w = do
+-- | Form wrapper which will automatically add a submit button to the form to handle submit on enter.
```
tomsmalley: This comment isn't true, you need to pass in a suitable submit button

alexfmpe: Oh right, when I updated the comment `form` was actually `setupForm`. Fixing

PR #301, opened 2019-12-03
Change default gas price and gas limit.

mankyKitty:
Update scaling function for transaction price slider.

The slider doesn't scale the gas price as smoothly as it did before because of the expanded range.

### DIFF
--- a/frontend/src/Frontend/Network.hs
+++ b/frontend/src/Frontend/Network.hs
@@ -390,12 +390,11 @@ getSelectedNetworkInfos networkL = do
     errNets <- sample $ current $ networkL ^. network_selectedNodes
     pure $ rights errNets
 
--- This is the minimum precision allowed by the Pact language:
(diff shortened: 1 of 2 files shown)


frontend/src/Frontend/Network.hs
```diff
@@ -404,10 +403,10 @@ defaultTransactionTTL = TTLSeconds (8 * 60 * 60) -- 8 hours
 
 -- Taken from https://github.com/kadena-io/chainweb-node/blob/85688ea0182d1b1ab0d8d784a48b4851a950ec7a/src/Chainweb/Chainweb.hs#L344
 chainwebGasLimit :: Num a => a
-chainwebGasLimit = 1e5
+chainwebGasLimit = 600
```
tomsmalley: This is still 100,000 (see upstream). This really should be renamed to `chainwebGasLimitMaximum` or something, to set it apart from the user configurable gas limit (which has the default from below)

tomsmalley (changes requested):

PR #307, opened 2019-12-03
Improve error and request key display for X-Chain Transfers

mankyKitty:
Errors in crosschain transfers are now displayed in a box that appears below the progress
icons. The box is cleared when the transfer is re-attempted.

The request key no longer displays inline and is display when it is available in a
separate box.

Some effort has been made to pretty-print the error information.

### DIFF
--- a/frontend/src/Frontend/UI/Dialogs/Send.hs
+++ b/frontend/src/Frontend/UI/Dialogs/Send.hs
@@ -58,6 +58,7 @@ import qualified Data.IntMap as IntMap
 import qualified Data.List as L
 import qualified Data.Map as Map
 import qualified Data.Text as T
+import qualified Data.Text.IO as T
 import qualified Pact.Server.ApiV1Client as Api
 import qualified Pact.Types.API as Api
 import qualified Pact.Types.Command as Pact
@@ -68,7 +69,6 @@ import qualified Servant.Client.JSaddle as S
 import qualified Text.URI as URI
 
 import Common.Wallet
-import Obelisk.Generated.Static
 import Frontend.Crypto.Class (HasCrypto)
 import Frontend.Foundation hiding (Arg)
 import Frontend.KadenaAddress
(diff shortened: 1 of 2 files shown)


frontend/src/Frontend/UI/Dialogs/Send.hs
```diff
@@ -542,15 +523,28 @@ finishCrossChainTransfer netInfo fromIndex fromAccount ucct toGasPayer = Workflo
...
-          el "p" $ crossChainMessage $ "Request key: " <> Pact.requestKeyToB16Text requestKey
+        item (pure Status_Done) $ el "p" $
+          text $ "Cross chain transfer initiated on chain " <> _chainId fromChain
         pb <- getPostBuild
         runUnfinishedCrossChainTransfer netInfo fromChain toChain toGasPayer $ requestKey <$ pb
+
+    elClass "h2" "heading heading_type_h2" $ text "Request Key"
+    divClass "group" $ text $ Pact.requestKeyToB16Text requestKey
+
+    void $ runWithReplace blank $ ffor (leftmost [Just <$> errMsg, Nothing <$ retry]) $ \case
```
alexfmpe, line 539: Maybe use `widgetHold_` since we're not consuming anything from `runWithReplace` anyway. 
Could also do `$ traverse $ \e ->`  instead of `\case` depending on your preferences

frontend/src/Frontend/UI/Dialogs/Send.hs
```diff
@@ -821,15 +822,22 @@ listenForContinuation
...
-      liftIO $ print es
-      pure $ Left $ tshow es
+    Left es -> packHttpErrors es
     Right (Api.ListenResponse cr) -> case Pact._crContinuation cr of
-      Nothing -> pure $ Left "Result was not a continuation"
-      Just pe -> pure $ Right (Pact._crReqKey cr, pe)
+      Nothing ->
+        pure $ Left $ T.unlines ["Result was not a continuation", tshow (Pact._crResult cr)]
+      Just pe ->
+        pure $ Right (Pact._crReqKey cr, pe)
```
alexfmpe, line 830: `pure` can be floated up one level

frontend/src/Frontend/UI/Dialogs/Send.hs
```diff
@@ -625,17 +619,30 @@ crossChainTransfer netInfo fromIndex fromAccount toAccount fromGasPayer crossCha
...
+
+    let errMsg = leftmost [keySetError, initiatedError, errMsg0]
+    void $ runWithReplace blank $ ffor (leftmost [Just <$> errMsg, Nothing <$ retry0]) $ \case
+      Just e -> do
+        elClass "h2" "heading heading_type_h2" $ text "Failure"
+        divClass "group" $ text e
+      Nothing -> do
+        blank
+
+    pure resultOk0
```
alexfmpe, line 645: This block seems very similar to the one above, maybe factor out some stuff?

PR #308, opened 2019-12-03
Return correct count of capabilities when they're static

mankyKitty:
The 'apply to all' will now appear correctly when capabilities are static.

### DIFF
--- a/frontend/src/Frontend/UI/DeploymentSettings.hs
+++ b/frontend/src/Frontend/UI/DeploymentSettings.hs
@@ -961,7 +961,7 @@ uiSenderCapabilities m cid mCaps mkSender = do
           rest <- staticCapabilityRows (Just <$> eApplyToAll) $ filter (not . isGas . _dappCap_cap) caps
           pure ( _capabilityInputRow_account gas
                , combineMaps [(_capabilityInputRow_value gas),rest]
-               , constDyn 0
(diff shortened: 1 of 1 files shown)


benkolera (approved):

frontend/src/Frontend/UI/DeploymentSettings.hs
```diff
@@ -961,7 +961,7 @@ uiSenderCapabilities m cid mCaps mkSender = do
           rest <- staticCapabilityRows (Just <$> eApplyToAll) $ filter (not . isGas . _dappCap_cap) caps
           pure ( _capabilityInputRow_account gas
                , combineMaps [(_capabilityInputRow_value gas),rest]
-               , constDyn 0
+               , constDyn (1 {- Gas payer -} + length caps)
                )
```
alexfmpe, line 965: Would the count be clearer/more robust if we did something like
```haskell
(rest, restCount) <- case mCaps of 
  (...)
pure ( _capabilityInputRow_account gas
     , combineMaps [(_capabilityInputRow_value gas), rest]
     , fmap (+1) (maybe 0 length <$> mcaps)
     )
```
?

mankyKitty: `mCaps` isn't a `Dynamic`, otherwise yeah that would be better.

PR #310, opened 2019-12-03
fix: Wrap the capabilities in the static signing screen.

benkolera:
I'm truly sorry for the three `!important`s. :sob:


tomsmalley:
Well, at least in this case it _is_ important :)

PR #317, opened 2019-12-03
Remove odd warning

alexfmpe:


frontend/src/Frontend/UI/Dialogs/AddVanityAccount.hs
```diff
@@ -199,10 +199,7 @@ vanityAccountCreateSubmit model dAccount chainId result nodeInfos = Workflow $ d
   pb <- getPostBuild
   resp <- performLocalRead (model ^. network) $ [req] <$ pb
   let localOk = fforMaybe resp $ \case
-        -- Generates a 'Pattern match has inaccessible right hand side' warning. Not sure why.
-        [(_, t)]
-          | Just (_, PLiteral (LString "Write succeeded")) <- justThat t
```
alexfmpe, line 204: @tomsmalley let me know if this was actually meant to be `justThere`

tomsmalley (commented):
Need an extra pattern for These too

frontend/src/Frontend/UI/Dialogs/AddVanityAccount.hs
tomsmalley: yup, still generates the warning though. I just pushed the extra case to this branch

mankyKitty (approved):

PR #330, opened 2019-12-06
Linux app

benkolera:
This both builds a nixos app and a debian. 

To run nixos:
`WEBKIT_DISABLE_COMPOSITING_MODE=1 $(nix-build -A linux)/bin/kadena-chainweaver` from the source

To run debian:
- In this source directory `nix-build -A deb`
- copy result/chainweaver_0.1.0_amd64.deb to an ubuntu machine. 
- On ubuntu machine terminal: 
  - `sudo dpkg -i kadena-chainweaver_0.1.0_amd64.deb`
  - Either run it from the application menu or run `kadena-chainweb-rc1` in a terminal if something is bad and you need to see some logs.

Current known issues across both nixos and debian:
- [x] Pact file not saved across executions of chainweaver
- [ ] Warnings about ArrayBuffer vs ArrayBufferView in the console
- [x] Doesn't output to syslog and just the console atm
- [x] Problems with compositing with nvidia and in virtual box. Best to disable
- [x] Writes to ~/.local/shared/chainweaver instead of .local/share. Typo

Current known issues for nixos:
- Doesn't disable compositing automatically in the wrapper script

Current known issues for debian:
- [x] First run is painfully slow (come up as not responding; may be my vm only has 4G mem)
- [x] External links don't open
- [x] File dialog crashes app
- [x] .deb could be smaller
- [x] Put static files in a better place than libexec
- [x] Signing server not running / responding
- [x] Needs to package up a pixbuf config and maybe fix gsettings provider too
- [x] Most likely want a desktop entry rather than running from a terminal :) 
- [x] With no CI / binary caching atm, the custom rebuild of webkitgtk is painful because it takes about 40 mins to build. 
- [ ] No testing on Mate (GTK 2), Cinnamon, XCFE distros. All testing has been done on Ubuntu 18 (GTK 3 and XOrg). Given that the deb is so self contained, it's probably more wayland that will mess us up rather than non-gtk3 distros. 

Alternative ways of going about this:
- Instead of copying the libraries across, statically link. Involves compiling things with musl instead which is probably more trouble than it's worth. Given that GTK is lgpl, its in our best interests not to statically link to it anyway. See: https://github.com/nh2/static-haskell-nix/issues/50
- Something like https://github.com/matthewbauer/nix-bundle where we package up a whole nix store . This produced a thing that was quite large and out of the box had a pretty slow startup time.


default.nix
```diff
@@ -103,6 +104,91 @@ in obApp // rec {
...
+      patchelf --set-rpath $LIBDIR $TMPEXE
+      cp $TMPEXE $BINDIR/${linuxAppName}
+
+      cp "${pkgs.z3}"/bin/z3 "$BINDIR/z3"
+      #TODO : This shouldn't be in libexec
+      cp -rL "${obApp.mkAssets obApp.passthru.staticFiles}" "$BINDIR/static.assets"
+      cp -rL "${obApp.passthru.staticFiles}" "$BINDIR/static"
+      cp -rL "${sass}/sass.css" "$BINDIR/sass.css"
+      ${pkgs.libicns}/bin/png2icns "$BINDIR/pact.icns" "${macAppIcon}"
+      ${pkgs.libicns}/bin/png2icns "$BINDIR/pact-document.icns" "${macPactDocumentIcon}"
```
alexfmpe: Don't know how linux app icons work, but I doubt they use Apple Icon Image format

default.nix
```diff
@@ -103,6 +104,203 @@ in obApp // rec {
...
+      mkdir -p $LIBDIR
+      mkdir -p $LIBEXECDIR
+
+      cp ${obApp.ghc.linux}/bin/linuxApp $TMPEXE
+      chmod u+w $TMPEXE
+       
+      cp $TMPEXE $CHAINWEAVER_LIBEXEC_EXE
+      cp "${pkgs.z3}"/bin/z3 "$LIBEXECDIR/z3"
+
+      #TODO : This shouldn't be in libexec
```
matthewbauer: Yeah, obelisk by default looks in executable directory for resources. You can use something like this to make it configurable via a cli flag:

- https://github.com/obsidiansystems/obelisk/pull/558
- https://gitlab.com/obsidian.systems/kiln/merge_requests/733/diffs#diff-content-5b1d1ec64e71b88e93f64a6e7dbb173c49239c6e

default.nix
```diff
@@ -103,6 +104,203 @@ in obApp // rec {
...
+
+      # Copy the webgtk libexec over because we really want the webworker. 
+      # This depends on our patch to ubuntuWebkitgtk for the lib to find the exec dir
+      cp ${ubuntuWebkitgtk}/libexec/webkit2gtk-4.0/* $LIBEXECDIR/
+      # We don't need the gi-repository folders in lib. They are a compile time gobject introspection
+ 
+      # Figure out how to make the array work with nix escaping
+      wrapperArgs=""
+      wrapperArgs+="--set WEBKIT_DISABLE_COMPOSITING_MODE 1 "
+      wrapperArgs+="--set LD_LIBRARY_PATH $LIBPATH "
```
matthewbauer: probably should be prefix to let others also override other things.

linux/src/Linux.hs
```diff
@@ -0,0 +1,183 @@
+{-# LANGUAGE OverloadedStrings #-}
+module Linux where
+
+-- import Control.Lens (_head, preview)
```
tomsmalley: leftover

linux/src/Linux.hs
```diff
@@ -0,0 +1,183 @@
...
+
+import Desktop (main', AppFFI(..))
+import WebKitGTK
+
+{-- NOTES
+
+To run: `WEBKIT_DISABLE_COMPOSITING_MODE=1 $(nix-build -A linux)/bin/kadena-chainweaver-rc1` from the project root
+
+TODO
+- Setting up app menu? What does the mac app do and what do we want in there?
```
tomsmalley: It just allows opening a file, quitting, and it also enables the copy/paste shortcuts. Probably not applicable here

linux/src/Linux.hs
```diff
@@ -0,0 +1,183 @@
...
+  Gtk.fileChooserSetSelectMultiple chooser False
+
+  res <- Gtk.nativeDialogRun chooser
+  void $ case toEnum (fromIntegral res) of
+    Gtk.ResponseTypeAccept ->  Gtk.fileChooserGetFilenames chooser >>= traverse_ handleOpen . headMay
+    _ -> pure ()
+
+moveToForeground :: Gtk.Window -> IO ()
+moveToForeground w = do
+  putStrLn "MoveToForeground"
```
tomsmalley: can these use logToSyslog?

tomsmalley (approved):
nitpicks only

linux/src/Linux.hs
benkolera: I don't think they are even useful anymore. I'll just nuke them. :)

PR #335, opened 2019-12-10
Remove 'Advanced' section from dialog for adding vanity account

alexfmpe:


frontend/src/Frontend/UI/DeploymentSettings.hs
```diff
@@ -532,13 +529,14 @@ uiCfg
   -> Maybe TTLSeconds
   -> Maybe GasLimit
   -> g (Text, m a)
+  -> Maybe (model -> Dynamic t Bool -> m (Event t (), ((), mConf)))
```
alexfmpe, line 532: This could easily be `[]` instead of `Maybe` since we just `<>` the `mConf` but the accordions are setup in a way that closing one opens the other, which only really works when there's 2

PR #337, opened 2019-12-10
Handle wallet closure or backend failures in vanity account creation

mankyKitty:


frontend/src/Frontend/UI/Dialogs/AddAccount.hs
```diff
@@ -78,19 +83,31 @@ uiCreateWalletStepOne
...
       dChainId <- divClass "group segment" $ do
         transactionDisplayNetwork model
-        userChainIdSelect model
+        userChainIdSelectWithPreselect model (fmap _account_chainId <$> dInflightAcc)
 
       elClass "h2" "heading heading_type_h2" $ text "Reference Data"
-      dNotes <- divClass "group segment" $
-        value <$> mkLabeledClsInput True "Notes" inpElem
+      dNotes <- divClass "group segment" $ fmap value $ mkLabeledClsInput True "Notes" $ inpElem $ def
+        & inputElementConfig_setValue .~ tagMaybe (fmap (unAccountNotes . _account_notes) <$> current dInflightAcc) pb
```
alexfmpe: If an inflight account is found *after* this dialog was opened, it seems to me that the above `dyn_` would make us claim the details were pre-populated, but at least this field won't react?

frontend/src/Frontend/UI/Dialogs/AddAccount.hs
```diff
@@ -107,15 +125,21 @@ uiCreateWalletStepOne model onClose = Workflow $ do
           pure (net, chain, mkAccountNotes notes)
         newConf = mempty & walletCfg_createWalletOnlyAccount .~ eAddAcc
 
+        eInflightFound = () <$ ffilter isNothing (updated dInflightAcc)
```
alexfmpe, line 128: I'd expect `isJust` ?

mankyKitty: I want this to fire only when when an inflight account has been 'found' on the chain, and therefore no-longer inflight so we have to jimmy up a new pair of keys.

So that would change the value in the dynamic from `Just account` to `Nothing`, so only fire on `Nothing`. I have to look up which direction the `filter` function is using the predicate *EVERY SINGLE TIME* ... so I wouldn't be surprised if I need to flip it.

PR #341, opened 2019-12-11
Allow launching receive dialog without a pre-specified chain

alexfmpe:
~On top of #340.~


frontend/src/Frontend/UI/Dialogs/Receive.hs
```diff
@@ -205,13 +208,18 @@ uiReceiveModal0 model account onClose = Workflow $ do
   let
     done = gate (current showingAddr) doneNext
     deploy = gate (not <$> current showingAddr) doneNext
-
-    infos = (liftA2 . liftA2) (,) netInfo transferInfo
+    ap2' = (liftA2 . liftA2) (&)
```
tomsmalley: This is a lot less readable than the old implementation imo

alexfmpe: I don't see how to keep the old implementation without lots of tupling and untupling and non-binary uncurrying.

frontend/src/Frontend/UI/Dialogs/Receive.hs
```diff
@@ -144,56 +145,58 @@ uiReceiveModal0
...
+  (showingAddr, chain, (conf, ttl, gaslimit, transferInfo)) <- divClass "modal__main receive" $ do
     rec
       showingKadenaAddress <- toggle True $ onAddrClick <> onReceiClick
 
-      (onAddrClick, _) <- controlledAccordionItem showingKadenaAddress mempty (text "Option 1: Copy and share Kadena Address")
+      elClass "h2" "heading heading_type_h2" $ text "Destination"
+      chain <- divClass "group" $ do
+        -- Network
+        void $ mkLabeledClsInput True "Network" $ \_ -> do
+          stat <- queryNetworkStatus (model ^. network_networks) $ pure $ _account_network account
```
mankyKitty: querying for the network status was removed in #338 and there was a function added for consistent modal headings. Might be worth rebasing `develop` ? There'll be conflicts, I imagine.

frontend/src/Frontend/UI/Dialogs/Receive.hs
tomsmalley: `push` and `sample` is probably the best bet here

tomsmalley: much clearer

PR #347, opened 2019-12-14
Convert exclusive 'These' to 'Either'

alexfmpe:


frontend/src/Frontend/ModuleExplorer/ModuleRef.hs
```diff
@@ -256,7 +255,7 @@ fetchModule networkL onReq = do
       deployedResultsZipped = first fst <$> deployedResults
 
     pure $ fmapMaybe listToMaybe . ffor deployedResultsZipped $ \(dmr, errs) ->
-      map ((dmr,) . (either This That . getModule . snd <=< first prettyPrintNetworkErrors)) errs
+      map ((dmr,) . (getModule . snd <=< first prettyPrintNetworkErrors . networkErrorResultToEither)) errs
```
alexfmpe, line 258: Actually, this is a semantic change, and probably wrong, because `These`'s `join` collects failures.

PR #348, opened 2019-12-15
Return error instead of empty module map when parsing/compiling fails

alexfmpe:


frontend/src/Frontend/ModuleExplorer/File.hs
```diff
@@ -206,11 +207,15 @@ fetchFile m onFileRef = do
...
-fileModules :: PactFile -> Map ModuleName (ModuleDef (Term Name))
+fileModules :: PactFile -> Either Text (Map ModuleName (ModuleDef (Term Name)))
 fileModules (Code c) = case Pact.compileExps (Pact.mkTextInfo c) <$> Pact.parseExprs c of
-  Right (Right terms) -> Map.fromList $ mapMaybe getModule terms
-  _                   -> Map.empty
+  Left err -> Left $ "Parsing failed: " <> tshow err
+  Right (Left err) -> Left $ "Compilation failed: " <> tshow err
+  Right (Right terms) -> Right $ Map.fromList $ mapMaybe getModule terms
 
+-- TODO: remove, only used for compatibility with older code
```
mankyKitty, line 216: Make a ticket for this so we don't forget?

alexfmpe: Done

PR #357, opened 2019-12-18
Tweak "Terms of Service" screen

alexfmpe:


backend/sass/common.blocks/_setup.scss
```diff
@@ -195,11 +191,9 @@ $kadena-orange: #F3AB3C;
     line-height: initial;
 
     &.button_type_confirm {
-      width: 96%;
```
mankyKitty, line 198: The loss of this width makes the 'Continue' button display incorrectly on the following screen. Compared `develop` to this branch with restarts.

PR #358, opened 2019-12-18
Prevent sidebar icons from twitching on hover

alexfmpe:


mankyKitty:
Logout button still twitches, but other icons are fixed.

alexfmpe:
Oops, was testing this in the Webapp version which doesn't have that button

PR #368, opened 2019-12-31
Schema Migrations

benkolera:
Okay, so this change comes in a few waves:

- Change storage to a Free Monad rather than a record of functions that is locked into JSM. We do thin because we want to test the crap out of the storage related functionality.
- Collapse all the separate frontend store keys into a single key so that we can centralise backups (we don't actually *need* to do this as this system would work with modular separate keys, but it would mean that the migration wouldn't be centralised at the toplevel UI like it is here and could easily be forgotten).
- Incur some additional constraints on our keys, namely GCompare, GShow, UniverseSome, AesonGADT and ArgDict. See caveats and hacks.
- Copy the old keys to Frontend.Store.V0 and put the new split accounts key in Frontend.Store.V1
- Write a backup to DMap function using the Universe instance to enumerate the key space.
- Write a restore from DMap function
- Write the upgrade process from the V0 DMap to the V1 Dmap then use the abstract dump and restore functions to power that.
- Write a boat load of tests

TODO:
- Testing on mac to make sure I haven't broken things.
- Better error handling in the migration. Right now the upgrade process forces the upgrade to be a total function from Dmap to Dmap, but it'd be great to be able to properly report aeson decode failures on the old version (right now things will just parse to Nothing if there is an error, which is the same mess that we had before).
- Some things could have better homes and more documentation. Esp the crazy abstract backup / restore stuff. That stuff is a little hairy, but also awesome (I think). It's very tricky to mess up.

Caveats and Hacks:
- The key types in Frontend.Store depend on unversioned datatypes. This could cause us pain later on, but there is a test in Frontend.StoreSpec that actually parses real storage JSON so if there are any breaking changes in those non versioned types then the tests ought to pick that up. At that point we can break out the types in old keys into simpler versioned types and make sure the tests pass. 
- Because everything is polymorphic to the private key type, nothing in this upgrade process can version that. Something else has to make sure that that type is backwards compat forever and nothing currently is in the test suite. 
- Data.GADT.{Show,Compare}.TH are both broken on our key type. So is UniverseSome. There are fixed versions of those TH functions in Frontend.Store.TH that should be upstreamed eventually.
- deriveArgDict is also broken, so we just hand splice in the generated code. Cale knows about this issue and has fixed it in AesonGADT but not in constraints-extras yet.
- All of these issues are to do with the extra key param in our type constructor.


frontend/src/Frontend/Store.hs
```diff
@@ -0,0 +1,76 @@
...
+  migratedE <- performEvent $ performUpdate <$ pbE
+  widgetHold_ blank $ widget <$ migratedE
+  where
+    performUpdate = do
+      _ <- runStorageJSM $ storageVersioner_upgrade v
+      pure ()
+
+versioner :: forall key. (ToJSON key, FromJSON key) => StorageVersioner (Latest.StoreFrontend key)
+versioner = StorageVersioner
+  { storageVersion_metaPrefix = prefix
```
tomsmalley:
```haskell
  { storageVersioner_metaPrefix = prefix
```

frontend/src/Frontend/Store/V0/Wallet.hs
```diff
@@ -0,0 +1,124 @@
...
+    pure $ KeyPair
+      { _keyPair_publicKey = public
+      , _keyPair_privateKey = private
+      }
+
+-- | We keep track of deletions at a given index so that we don't regenerate
+-- keys with BIP32.
+data SomeAccount key
+  = SomeAccount_Deleted
+  | SomeAccount_Inflight (Account key)
```
tomsmalley: Hmm I think this is the wrong version of the code - the released version doesn't have inflight stuff iirc
https://github.com/kadena-io/chainweaver/tree/beta-1

common/src/Common/Wallet.hs
```diff
@@ -678,6 +680,17 @@ accountUnfinishedCrossChainTransfer = view (accountInfo . accountInfo_unfinished
 accountBalance :: Account -> Maybe AccountBalance
 accountBalance = view (accountInfo . accountInfo_balance)
 
+upsertNonVanityBalance :: NetworkName -> PublicKey -> ChainId -> Maybe AccountBalance -> AccountStorage -> AccountStorage
+upsertNonVanityBalance net pk chain balance store = store
+  & (_AccountStorage . at net %~ (Just . maybe mempty id))
+  & (_AccountStorage . ix net . accounts_nonVanity . at pk %~ (Just . maybe mempty id))
```
mankyKitty: `(Just . fold)` instead of `(Just . maybe mempty id)` ?

frontend/src/Frontend/Store/V1.hs
```diff
@@ -0,0 +1,177 @@
...
+          else Accounts
+            mempty
+            (Map.singleton newPubKey (Map.singleton newChainId (NonVanityAccount newAccountInfo)))
+
+      in AccountStorage $ Map.singleton (V0._account_network a) accounts
+
+    upgradePublicKey = PublicKey . V0.unPublicKey
+
+    upgradeAccountNotes a = mkAccountNotes (V0.unAccountNotes . V0._account_notes $ a)
+    --This is a bit unfortunate
```
tomsmalley, line 132: Yeah, we need to regenerate these.

tomsmalley (changes requested):

frontend/src/Frontend/Store.hs
tomsmalley: I also changed this to add the prefixed underscore.

common/src/Common/Wallet.hs
benkolera: Oh yeah, good point. :)

frontend/src/Frontend/Storage/Class.hs
```diff
@@ -0,0 +1,299 @@
...
+   want to export the whole store because I don't think that we want to export the
+   encrypted keys. It's not quite what we want for the versioning process so we are not
+   going to get it "for free" sadly.
+ * The current storage type locks things up in JSM which makes it really tricky to test.
+   We could run the frontend storage tests in ghcjs but then we'd be stuck not being able to
+   test the desktop migrations. We could parameterise Storage with a type parameter for the
+   inner monad, but then HasStorage needs to become a MTPC and stuff gets weird for the rest
+   of the app. Really, we just want to be able to test the derived functions from the algebra
+   and we don't want "testability" stuff bleeding out into the app.
+   For this reason, I'm going to change Storage to a Free Monad and HasStorage / StorageT just
```
alexfmpe, line 66: Not so free anymore

common/src/Common/GistStore.hs
```diff
@@ -0,0 +1,22 @@
...
+-- | Meta data about a gist. (Like name and description)
+data GistMeta = GistMeta
+  { _gistMeta_fileName    :: Text
+    -- ^ What filename to use in the gist.
+  , _gistMeta_description :: Text
+    -- ^ The description of the gist to create.
+  }
+  deriving (Show, Generic)
+
+-- TODO Remove these derived instances
```
alexfmpe, line 18: Why? When?

frontend/src/Frontend/Store/V1.hs
```diff
@@ -0,0 +1,176 @@
...
+import Common.Network (NetworkName, NodeRef)
+import Common.OAuth (OAuthProvider(..))
+import Common.GistStore (GistMeta)
+
+import Frontend.Store.TH
+import qualified Frontend.Store.V0 as V0
+import qualified Frontend.Store.V0.Wallet as V0
+import Frontend.Store.MigrationUtils
+
+-- WARNING: Upstream deps. Check this when we bump pact and obelisk!
```
alexfmpe, line 31: Tests will catch breakage from upstream bumps right?

frontend/src/Frontend/Wallet.hs
```diff
@@ -224,31 +225,49 @@ getBalances
...
+      -- accounts on generateKey because if we just did it on create then we'd
+      -- get the user stuck if they ever added in a new network.
+      let
+        chainsToAccounts = fold
+            [ foldMap
+              (\(an, m) ->
+                foldMap (\cId ->
+                  MonoidalMap.singleton cId (Set.singleton (unAccountName an)))
+                (Map.keys m))
+              (mAccounts ^.. _Just . accounts_vanity . to Map.toList . traverse)
```
alexfmpe, line 256: How about 
``` haskell
fold
  [ flip foldMap mAccounts $ foldAccounts $ \case
    AccountRef_NonVanity _ _ :=> _ -> mempty
    AccountRef_Vanity an cId :=> _ -> MonoidalMap.singleton cId (Set.singleton (unAccountName an))
  ,
```

common/src/Common/GistStore.hs
benkolera: Because the derived instances are fragile to making incompatible changes to the json structures accidentally. We have a policy now to not derive json instances.

It probably should have already been done, but gists don't work on desktop anyway.

frontend/src/Frontend/Storage/Class.hs
benkolera: True. We could probably just delete this comment all together now.

frontend/src/Frontend/Store/V1.hs
benkolera: They should do, yeah.

frontend/src/Frontend/Wallet.hs
benkolera: Fold accounts didn't work here anymore, but I forget why. I'll try to remember. :)

alexfmpe: It typechecks but can be much less efficient since it visits all the non-vanity accounts for what's basically a no-op, which could probably be fixed with 
```haskell
flip foldMap mAccounts $ \a -> flip foldAccounts (a & accounts_vanity .~ mempty) $ \case
```
but at that point the readability gain is somewhat lost.

Let's leave it as is. If this usecase starts showing up a lot we can just break `foldAccounts` up into vanity and nonvanity.

common/src/Common/GistStore.hs
alexfmpe: Ah nevermind, I thought the comment also applied to `Semigroup`.

PR #372, opened 2020-01-03
Remove the chain table from the key details dialog. 

mankyKitty:
Fix an issue where the horizontal separator would display incorrectly on the account details dialog.

### DIFF
--- a/frontend/src/Frontend/UI/Dialogs/AccountDetails.hs
+++ b/frontend/src/Frontend/UI/Dialogs/AccountDetails.hs
@@ -90,12 +90,14 @@ uiAccountDetailsDetails netname a onClose = Workflow $ do
       -- Notes edit
       notesEdit0 :: Maybe (Dynamic t Text) <- case accountNotes a of
         -- Only vanity accounts have notes.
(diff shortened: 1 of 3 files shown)


frontend/src/Frontend/UI/Dialogs/AccountDetails.hs
```diff
@@ -90,12 +90,14 @@ uiAccountDetailsDetails netname a onClose = Workflow $ do
...
+        Just va -> do
+          notes0 <- fmap (Just . value) $ mkLabeledClsInput False "Notes" $ \cls -> uiInputElement $ def
+            & inputElementConfig_initialValue .~ unAccountNotes va
+            & initialAttributes . at "class" %~ pure . maybe (renderClass cls) (mappend (" " <> renderClass cls))
+          horizontalDashedSeparator
+          pure notes0
         _ -> pure Nothing
-      -- separator
-      horizontalDashedSeparator
+        -- separator
```
alexfmpe: Leftover

mankyKitty: Moved this to its rightful place.

PR #376, opened 2020-01-06
Centralise the logging & remove all `putStrLn` calls

mankyKitty:
Replaced some to get a feel for the utility of it.

Gathering some feedback before I do the chore of replacing all the calls to `putStrLn` and friends.

### DIFF
--- a/frontend/src/Frontend/Log.hs
+++ b/frontend/src/Frontend/Log.hs
@@ -8,7 +8,7 @@ module Frontend.Log
   , formatLogMessage
   , defaultLogger
   , logOn
-  , logPromptly
+  , putLog
   , fmtLogTH
   ) where
 
@@ -36,7 +36,7 @@ makePactLenses ''LogCfg
 
 data Logger t = Logger
   { _log_formatMessage :: LogLevel -> Text -> LogStr
-  , _log_logPromptly :: LogLevel -> Text -> IO ()
+  , _log_putLog :: LogLevel -> Text -> IO ()
   }
 
 makePactLenses ''Logger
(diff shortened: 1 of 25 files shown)


frontend/src/Frontend/Wallet.hs
```diff
@@ -270,23 +272,24 @@ getBalances model accStore = performEventAsync $ flip push accStore $ \(keys, ne
 
       -- Build a request for each chain
       requests <- flip MonoidalMap.traverseWithKey chainsToAccounts $ \chain as -> do
-        liftIO $ putStrLn $ "getBalances: Building request for get-balance on chain " <> T.unpack (_chainId chain)
+        logStrPromptly model . $(fmtLogTH LevelInfo) $ "getBalances: Building request for get-balance on chain " <> _chainId chain
```
benkolera: Is promptly the right word here? It has pretty strong connotations in the reflex space.

frontend/src/Frontend/Log.hs
```diff
@@ -0,0 +1,83 @@
...
+newtype LogCfg t = LogCfg
+  { _logCfg_logMessage :: Event t (LogLevel, Text)
+  }
+
+makePactLenses ''LogCfg
+
+data Logger t = Logger
+  { _log_formatMessage :: LogLevel -> Text -> LogStr
+  , _log_logPromptly :: LogLevel -> Text -> IO ()
+  , _log_logStr :: LogStr -> IO ()
```
benkolera: I'm thinking this shouldn't exist so that we always force folk to choose the log level so that we could filter out certain log levels later? And then if you don't have it you probably don't need formatMessage.

frontend/src/Frontend/Wallet.hs
mankyKitty: It's not but it was the first thing that came to mind as I wanted to differentiate it from the `Event t (LogLevel, Text)`

frontend/src/Frontend/Log.hs
mankyKitty: Removed this along with the TH stuff that was a bit too janky.

benkolera (approved):

mac/src/Mac.hs
```diff
@@ -34,8 +39,26 @@ ffi = AppFFI
   , _appFFI_resizeWindow = uncurry resizeWindow
   , _appFFI_global_openFileDialog = global_openFileDialog
   , _appFFI_global_getStorageDirectory = getStorageDirectory
+  , _appFFI_global_logFunction = logToSyslog
   }
 
+logToSyslog :: LogLevel -> LogStr -> IO ()
```
alexfmpe, line 47: Seems equal to the `logToSyslog` in linux. Shouldn't this be in `desktop` ?

PR #380, opened 2020-01-07
Remove the requirement for a gas payer to be selected.

mankyKitty:
Add the "transaction sender" field to the transaction configuration pane and replace the uses of gas payer.

If a gas payer is selected then the their balance is checked and the transaction will fail if they have insufficient gas.

Added a notice that is displayed on the preview tab that advises the user that no gas payer has been selected.

### DIFF
--- a/frontend/src/Frontend/UI/DeploymentSettings.hs
+++ b/frontend/src/Frontend/UI/DeploymentSettings.hs
@@ -8,7 +8,6 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE PatternGuards #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE RecursiveDo #-}
 {-# LANGUAGE ScopedTypeVariables #-}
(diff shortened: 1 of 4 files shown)


frontend/src/Frontend/UI/DeploymentSettings.hs
```diff
@@ -251,14 +252,19 @@ buildDeploymentSettingsResult m mSender signers cChainId capabilities ttl gasLim
...
-  for_ (lookupAccountBalance sender allAccounts) $ \case
-    Nothing -> empty
-    Just b -> let GasLimit lim = _pmGasLimit publicMeta
-                  GasPrice (ParsedDecimal price) = _pmGasPrice publicMeta
-               in guard $ unAccountBalance b > fromIntegral lim * price
+  gasPayer <- lift mGasPayer
+  case gasPayer of
+    Nothing -> pure () -- No gas payer selected, move along
+    Just gp ->  for_ (lookupAccountBalance gp allAccounts) $ \case
+      Nothing -> empty -- Gas Payer selected but they're not an account?!
```
alexfmpe: Should we show some UI feedback in this case?

frontend/src/Frontend/UI/DeploymentSettings.hs
```diff
@@ -1002,12 +1028,22 @@ uiDeployPreview model settings keys accounts signers gasLimit ttl code lastPubli
   where
     uiPreviewResponses signing (cmd, wrappedCmd) = do
       pb <- getPostBuild
+
+      unless (getAny $ foldMap (Any . any isGas) capabilities) $ do
```
alexfmpe: Why not `any (any isGas) capabilities` ?

mankyKitty: Why not indeed!

frontend/src/Frontend/UI/DeploymentSettings.hs
mankyKitty: Added some feedback and refactored this function to provide verbose error information if more is required in future.

PR #383, opened 2020-01-08
Allow signing arbitrary texts in key details dialog

alexfmpe:


backend/sass/common.blocks/_wallet.scss
```diff
@@ -83,3 +83,27 @@
...
+  justify-content: space-between;
+
+  background: $chainweaver-blue;
+  color: white;
+
+  .copy-status {
+    transition: opacity, transform 0.25s linear;
+  }
+
+  .copy-success {
```
mankyKitty: Should this include the `&` : `&.copy-success` ?

alexfmpe: Hmm not like this, but it should if they're nested inside `.copy-status` like the setup ones are.
I'll nest here as well

frontend/src/Frontend/UI/Button.hs
```diff
@@ -192,6 +196,27 @@ copyButton cfg t = do
     _ <- copyToClipboard $ tag t onClick
     pure onClick
 
+copyButtonStatus :: (DomBuilder t m, MonadFix m, MonadHold t m, MonadJSM (Performable m), PerformEvent t m, PostBuild t m)
+                 => UiButtonCfg
```
tomsmalley: cursed indentation

frontend/src/Frontend/UI/Dialogs/KeyDetails.hs
```diff
@@ -62,15 +69,50 @@ uiKeyDetailsDetails keyIndex key onClose onCloseExternal = Workflow $ do
...
-      fmap value $ mkLabeledClsInput False "Notes" $ \cls -> uiInputElement $ def
+      notes <- fmap value $ mkLabeledClsInput False "Notes" $ \cls -> uiInputElement $ def
         & inputElementConfig_initialValue .~ unAccountNotes (_key_notes key)
         & initialAttributes . at "class" %~ pure . maybe (renderClass cls) (mappend (" " <> renderClass cls))
 
+      void $ accordionItemWithClick False mempty (accordionHeaderBtn "Advanced") $ withSecretKey $ \pk -> do
+        txt <- fmap value $ mkLabeledClsInput False "Text to sign" $ \cls -> uiTextAreaElement $ def
+          & initialAttributes .~ "class" =: renderClass cls
+
+        sigEv <- do
```
tomsmalley: this do block seems pointless

alexfmpe: Ah yes

frontend/src/Frontend/UI/Button.hs
alexfmpe: Thank you for the exorcism

PR #384, opened 2020-01-08
Redo public key truncation

mankyKitty:
New format is the first 6 characters, `...`, then the last 6 characters of the public key.

Added function in `Widgets` for when it is needed else where.

### DIFF
--- a/frontend/src/Frontend/UI/Widgets.hs
+++ b/frontend/src/Frontend/UI/Widgets.hs
@@ -51,6 +51,7 @@ module Frontend.UI.Widgets
   , uiAccountBalance'
   , uiAccountChain
   , uiAccountNotes
+  , uiPublicKeyShrunk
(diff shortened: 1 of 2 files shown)


frontend/src/Frontend/UI/Widgets.hs
```diff
@@ -586,6 +587,11 @@ uiAccountChain = _chainId . accountChain
 uiAccountNotes :: Account -> Text
 uiAccountNotes = maybe "" unAccountNotes . accountNotes
 
+uiPublicKeyShrunk :: PublicKey -> Text
+uiPublicKeyShrunk pk = (T.take 6 ktxt) <> "..." <> (T.drop (T.length ktxt - 6) ktxt)
```
tomsmalley: `T.takeEnd 6`

benkolera (approved):

PR #393, opened 2020-01-09
fix: Wallet model only returns non hidden keys/accounts

benkolera:
Also fixes the fake key to actually make a key of 64 0s properly and hides all accounts of a key when the key is hidden.


benkolera:
A big note on this PR. Before this PR, there is nothing that is hiding accounts that are attached to a key when a key is hidden. We should fix that. I'm happy to put that in here or a separate PR pending review of this PR.

PR #396, opened 2020-01-09
Make public key squishable

tomsmalley:


frontend/src/Frontend/UI/Widgets.hs
```diff
@@ -588,10 +588,13 @@ uiAccountChain = _chainId . accountChain
 uiAccountNotes :: Account -> Text
 uiAccountNotes = maybe "" unAccountNotes . accountNotes
 
-uiPublicKeyShrunk :: PublicKey -> Text
-uiPublicKeyShrunk pk = (T.take 6 ktxt) <> "..." <> (T.takeEnd 6 ktxt)
+uiPublicKeyShrunk :: (DomBuilder t m, PostBuild t m) => Dynamic t PublicKey -> m ()
+uiPublicKeyShrunk pk = do
+  divClass "wallet__public-key" $ do
+    elClass "span" "wallet__public-key__prefix" $ dynText $ T.dropEnd 6 <$> ktxt
```
benkolera: Why do you drop stuff from the prefix if it's hidden by overflow and the goal (iiuc) is to allow the user to double click the left bit and copy the whole key?

benkolera (approved):

PR #398, opened 2020-01-10
Default gas payer to be selected sender on vanity account creation

mankyKitty:
Prevent progress in vanity account creation unless gas payer is selected


frontend/src/Frontend/UI/Dialogs/AddVanityAccount.hs
```diff
@@ -144,7 +144,8 @@ uiAddVanityAccountSettings ideL onInflightChange mInflightAcc mChainId initialNo
           $ uiSenderDropdown def never
 
       (mGasPayer, signers, capabilities) <- tabPane mempty curSelection DeploymentSettingsView_Keys $
-        uiSenderCapabilities ideL cChainId Nothing mSender $ uiSenderDropdown def never ideL cChainId
+        uiSenderCapabilities ideL cChainId Nothing mSender
+          $ uiSenderDropdown def (updated mSender) ideL cChainId
```
benkolera: If the user goes back a page and changes the sender, wont this trash what the user previously chose? Is that OK?

mankyKitty: - Yes
- Not really, no

benkolera (approved):

PR #414, opened 2020-01-13
Don't ask for pass after completing setup

alexfmpe:


desktop/src/Desktop/Setup.hs
```diff
@@ -128,7 +128,7 @@ kadenaWalletLogo = divClass "logo" $ do
   elClass "div" "chainweaver" $ text "Chainweaver"
   elClass "div" "by-kadena" $ text "by Kadena"
 
-type SetupWF t m = Workflow t m (WalletScreen, Event t Crypto.XPrv)
+type SetupWF t m = Workflow t m (WalletScreen, Event t (Crypto.XPrv, Text))
```
mankyKitty: Should we `newtype` this? It's only used in this file, so not sure it's strictly necessary.

alexfmpe: Heh, it's fine either way. Personally I'd rather not have `Workflow t m` inside the newtype

mankyKitty: Sorry, was referring to the `Text` :)

alexfmpe: Oh that sounds good, that way we can not have a Show instance.

PR #416, opened 2020-01-13
Only color a setup step as blue in the header after leaving its screen

alexfmpe:


desktop/src/Desktop/Setup.hs
```diff
@@ -163,15 +163,15 @@ walletSetupRecoverHeader currentScreen = setupDiv "workflow-header" $ do
 
     isActive sid = sid `elem` (progress currentScreen)
     progress WalletScreen_Password =
```
alexfmpe, line 165: This would better be represented with a partial order, but `lattices` is giving me cabal hell right now

alexfmpe: Actually, nevermind, even defining `leq` seems too boilerplate-ish.

PR #420, opened 2020-01-14
OVA creation stuff

benkolera:
This creates an OVA with chainweaver inside that is should be runnable on all machines that can run virtualbox. 

It sets up an automatic port forward to the signing api port so it should be a zero config setup aside from the appliance import.


linux.nix
```diff
@@ -25,6 +26,15 @@
       ln -s "${sass}/sass.css" "$out/bin/sass.css"
     '';
   });
+  nixosDesktopItem = pkgs.makeDesktopItem {
+     name = linuxAppName;
+     desktopName = appName;
+     exec = "${nixosExe}/bin/${linuxAppName}";
+     icon= linuxAppIcon;
```
alexfmpe: Missing space

README.md
```diff
@@ -18,6 +18,24 @@ Run [`./scripts/devWeb`](./scripts/devWeb)
...
+
+### OVA Release
+
+The OVA includes an upgrade script inside the VM that allows the user to upgrade the nixos configuration without redownloading the whole OVA.
+
+This is powered by the http://nixcache.kadena.io binary cache and a file in s3 that contains the latest nixos system store path. See https://chainweaver-builds.s3.amazonaws.com/vm/master-store-path
+
+To release a new version, be sure that CI is pushing to that binary cache, and when the build passes QA overwrite the contents of master-store-path with the output of `nix-build -A chainweaverVMSystem default.nix` by running `scripts/update-vm-master-store`. This requires permission to write to the chainweaver-builds bucket and to have credentials in ~/.aws/credentials already.
+
+At this point, users ought to be able to run `update-chainweaver` from a terminal to switch to the new system config.
```
alexfmpe:
```suggestion
At this point, users ought to be able to run `upgrade-chainweaver` from a terminal to switch to the new system config.
```

virtualbox-image.nix
```diff
@@ -0,0 +1,117 @@
+{ config, lib, pkgs, ... }:
+
+# This is copied from nixpkgs just with the following changes:
```
alexfmpe, line 3: Is there a way to 'import' and override with these changes? Is it worth it?

benkolera: Nah. It's a big blob of shell script that is the thing that is being overridden. 

Ideally, the nixpkgs one would be more configurable and we'd upstream it.

PR #421, opened 2020-01-14
Unify copy buttons

alexfmpe:


mankyKitty:
This change means that the copy icon is no longer visible on hover because loading the svg as image means we can't affect the colour. Not sure how much of an issue this is ?

alexfmpe:
That is currently the case for the copy button in the passphrase screen, but zeplin shows the icon with the same color as the text in the dialog copy buttons, so I set it to white and now it's visible on hover.

PR #423, opened 2020-01-14
Add inline error reporting widget.

mankyKitty:
Created widget that does some error capturing and reports the given feedback inline in the
input element.

Adjusted affected fields.

Rebuilt the kadena address and account name inputs to use the new widget. Removed
adjustments that made those fields sit at a different spacing to other fields.

Added spacing to the Private Key field on the Receive dialog so it now matches other fields.

### DIFF
--- a/frontend/src/Frontend/UI/Widgets.hs
+++ b/frontend/src/Frontend/UI/Widgets.hs
@@ -73,7 +73,7 @@ module Frontend.UI.Widgets
   , noAutofillAttrs
   , addNoAutofillAttrs
   , horizontalDashedSeparator
-  , inlineInputFeedbackWrapper
+  , dimensionalInputFeedbackWrapper
   , uiSidebarIcon
   ) where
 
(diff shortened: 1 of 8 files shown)


frontend/src/Frontend/UI/Widgets.hs
```diff
@@ -700,9 +729,9 @@ paginationWidget cls currentPage totalPages = elKlass "div" (cls <> "pagination"
 horizontalDashedSeparator :: DomBuilder t m => m ()
 horizontalDashedSeparator = divClass "horizontal-dashed-separator" blank
 
-dimensionalInputWrapper :: DomBuilder t m => Text -> m a -> m a
-dimensionalInputWrapper units inp = divClass "dimensional-input-wrapper" $ do
-  divClass "dimensional-input-wrapper__units" $ text units
+inlineInputFeedbackWrapper ::  DomBuilder t m => Maybe Text -> m a -> m a
+inlineInputFeedbackWrapper units inp = divClass "dimensional-input-wrapper" $ do
```
alexfmpe: This class name is outside of the `traverse _ units` so it now feels off

PR #424, opened 2020-01-15
First cut of Kadena address overhaul

mightybyte:


mankyKitty:
Will need a bit more work after #423 is merged as it involves validation of the kadena address field.

tomsmalley:
Closing due to inclusion in #432

PR #429, opened 2020-01-16
Limit display of home button for Module Explorer

mankyKitty:
Only show the home button in the module explorer when the current module
is not the top of the stack.


frontend/src/Frontend/UI/ModuleExplorer/ModuleDetails.hs
```diff
@@ -74,10 +74,20 @@ moduleDetails
...
+        hb <- el "div" $ do
+          onHome <- if currentStackSize <= 1 then
+              pure never
+            else
+              homeButton "heading__left-double-button"
+
+          onBack <- backButton
+          pure (onHome, onBack)
+
+          -- (,) <$> homeButton "heading__left-double-button" <*> backButton
```
alexfmpe: Leftover comment

PR #440, opened 2020-01-22
Use /info instead of parsing swagger.json

benkolera:
Will the chains ever be non sequential? Wonder whether it's worth parsing the list of chains into there or whether the count is all we'll ever need?


tomsmalley:
Code looks good, but good question. Maybe we should just stick the list into the `ChainwebInfo` type instead.

PR #443, opened 2020-01-22
WIP: Always parse certificate with LibreSSL

alexfmpe:
From https://github.com/reflex-frp/reflex-platform/pull/625
Fails to build though, probably fixed in the newer nixpkgs current reflex-platform develop is using


alexfmpe:
Currently pinned reflex-platform includes this

PR #447, opened 2020-01-23
Add new widgets for 'Define Keyset' for Create accounts

mankyKitty:


frontend/src/Frontend/UI/Dialogs/AddVanityAccount.hs
```diff
@@ -173,20 +178,103 @@ createAccountNotGasPayer name chain keyset = Workflow $ do
...
+  :: ( MonadWidget t m
+     , HasWallet model key t
+     , HasJsonData model t
+     )
+  => model
+  -> m (Dynamic t (Maybe AddressKeyset))
 defineKeyset model = do
-  let allKeys = fmap (_keyPair_publicKey . _key_pair) . IntMap.elems <$> model ^. wallet_keys
+  let
+    allPreds = model ^. jsonData_keysets >>= fmap catMaybes . sequenceA . fmap _keyset_pred . Map.elems
```
alexfmpe: `joinDynThroughMap`

frontend/src/Frontend/UI/Dialogs/AddVanityAccount.hs
```diff
@@ -173,20 +178,103 @@ createAccountNotGasPayer name chain keyset = Workflow $ do
...
+      <$> model ^. wallet_keys
+
+    uiSelectKey = mkLabeledClsInput False (constDyn T.empty) $ const
+      $ uiDropdown 0 ((Map.fromList . IntMap.toAscList) <$> dAllKeys) $ def
+      & dropdownConfig_attributes .~ constDyn ("class" =: "labeled-input__input")
+
+    toIntSet :: IntMap.IntMap (Dropdown t IntMap.Key) -> Dynamic t (IntSet.IntSet)
+    toIntSet = foldMap (fmap adjustForSelectKey . value)
+      where
+        adjustForSelectKey k =
```
alexfmpe: This seems kind of fragile. I'd try `Maybe Int` because that would force us to handle the "Select" diferently than valid keys at every turn.

frontend/src/Frontend/UI/Dialogs/AddVanityAccount.hs
```diff
@@ -173,20 +178,103 @@ createAccountNotGasPayer name chain keyset = Workflow $ do
...
+          if k /= selectMsgKey then
+            -- Remove the adjustment for having a placeholder "Select" key
+            IntSet.singleton (k - 1)
+          else
+            IntSet.empty
+
+  dSelectedKeys <- uiAdditiveInput (const uiSelectKey) _dropdown_change (/= selectMsgKey) (== selectMsgKey)
+
+  -- TODO make this less awful
```
alexfmpe: How could this be improved?

frontend/src/Frontend/UI/Dialogs/AddVanityAccount/DefineKeyset.hs
```diff
@@ -0,0 +1,185 @@
...
+
+emptyKeysetPresets :: forall t. Reflex t => DefinedKeyset t
+emptyKeysetPresets = DefinedKeyset
+  { _definedKeyset_internalKeys = KeysetInputs memptyDyn memptyDyn
+  , _definedKeyset_externalKeys = KeysetInputs memptyDyn memptyDyn
+  , _definedKeyset_predicate = memptyDyn
+  }
+  where
+    memptyDyn :: Monoid m => Dynamic t m
+    memptyDyn = constDyn mempty
```
alexfmpe: Can just use `mempty :: Dynamic t m`

frontend/src/Frontend/UI/Dialogs/AddVanityAccount.hs
mankyKitty: I was hoping to be able to improve on all the flipping and rebuilding of `IntMap`/`Set`/`Map`/`IntSet`. Not sure how without rebuilding the whole thing.

frontend/src/Frontend/UI/Dialogs/AddVanityAccount/DefineKeyset.hs
```diff
@@ -0,0 +1,173 @@
...
+
+    externalKeys <- mkLabeledClsInput False "External Keys" $ const
+      $ uiExternalKeyInput $ _keysetInputs_value $ _definedKeyset_externalKeys presets
+
+    predicate <- mkLabeledClsInput False "Predicate (Keys Required to Sign for Account)" $ const
+      $ fmap value $ uiDropdown mempty allPredSelectMap $ def
+      & dropdownConfig_attributes .~ constDyn ("class" =: "labeled-input__input")
+      & dropdownConfig_setValue .~ (current (_definedKeyset_predicate presets) <@ pb)
+
+  -- TODO validate this (?? validation ??)
```
tomsmalley, line 166: This comment may be obsolete now. I originally made it for the freeform text predicate

PR #449, opened 2020-01-23
Add keyset column to accounts table

tomsmalley:


frontend/src/Frontend/UI/Wallet.hs
```diff
@@ -245,6 +253,13 @@ uiAccountItem keys name accountInfo = do
       let balance = (^? account_status . _AccountStatus_Exists . accountDetails_balance) <$> dAccount
       pure (balance, action)
 
+keysetSummary :: AddressKeyset -> Text
+keysetSummary ks = T.intercalate ", "
+  [ tshow numKeys <> if numKeys == 1 then " key" else " keys"
+  , _addressKeyset_pred ks
+  , "[" <> T.intercalate ", " (fmap keyToText . Set.toList $ _addressKeyset_keys ks) <> "]"
```
alexfmpe, line 260: The designs hint at a list of keys, but the ticket says to only show the first one?

tomsmalley: Oh, I was going off the design. Will have to clarify, but there’d be no noticeable difference

PR #456, opened 2020-01-24
Add 'Max' button to gas field

alexfmpe:
On top of #487


frontend/src/Frontend/UI/Dialogs/Send.hs
```diff
@@ -335,48 +335,62 @@ sendConfig model initData = Workflow $ do
...
 
-          (_, amount, _) <- mkLabeledInput True "Amount" uiGasPriceInputField
-            (def & inputElementConfig_initialValue .~ (maybe "" tshow mInitAmount))
-            -- We're only interested in the decimal of the gas price
-            <&> over (_2 . mapped . mapped) (\(GasPrice (ParsedDecimal d)) -> d)
+          let
+            balance = _account_status fromAcc ^? _AccountStatus_Exists . accountDetails_balance . to unAccountBalance
+
+            gasInputWithMaxButton cfg = mdo
+              let projectAmount = over (_2 . mapped . mapped) (\(GasPrice (ParsedDecimal d)) -> d)
```
benkolera: Could you not get the Wrapped instance to work either? It was doing weird things for me.

frontend/src/Frontend/UI/Dialogs/Send.hs
```diff
@@ -335,48 +335,62 @@ sendConfig model initData = Workflow $ do
...
+                  attrs = ffor useEntireBalance $ \u -> "disabled" =: ("disabled" <$ u)
```
benkolera: There is no real style change on the disabled. Probably good to add some CSS to make it look proper disabled, because it looks confusing to me in brave. Or, display the actual balance in the field when max is toggled.

frontend/src/Frontend/UI/Dialogs/Send.hs
```diff
@@ -335,48 +335,62 @@ sendConfig model initData = Workflow $ do
...
+          pure (validatedTxBuilder, validatedAmount, useEntireBalance)
 
         dialogSectionHeading mempty  "Transaction Settings"
-        (conf, _, _) <- divClass "group" $ uiMetaData model Nothing Nothing
-        pure (conf, recipient)
+        (conf, _, _, _) <- divClass "group" $ uiMetaData model Nothing Nothing
+        pure (conf, useEntireBalance, txBuilder, amount)
       mCaps <- tabPane mempty currentTab SendModalTab_Sign $ do
-        dedrecipient <- eitherDyn $ (fmap . fmap) fst recipient
+        dedrecipient <- eitherDyn txBuilder
```
benkolera, line 390: Is it worth putting feedback on this tab that the gas payer can't be the sender because max is clicked?

frontend/src/Frontend/UI/Dialogs/Send.hs
```diff
@@ -335,48 +335,62 @@ sendConfig model initData = Workflow $ do
...
+            balance = _account_status fromAcc ^? _AccountStatus_Exists . accountDetails_balance . to unAccountBalance
+
+            gasInputWithMaxButton cfg = mdo
+              let projectAmount = over (_2 . mapped . mapped) (\(GasPrice (ParsedDecimal d)) -> d)
+                  attrs = ffor useEntireBalance $ \u -> "disabled" =: ("disabled" <$ u)
+              (_, inputAmount, _) <- fmap projectAmount $ uiGasPriceInputField $ cfg
+                & inputElementConfig_elementConfig . elementConfig_modifyAttributes .~ updated attrs
+              useEntireBalance <- case balance of
+                Nothing -> pure $ pure Nothing
+                Just b -> fmap (\v -> b <$ guard v) . _checkbox_value <$> uiCheckbox "input-max-toggle" False def (text "Max")
```
benkolera: Obviously go by the designs we were given, but it really feels like this should be called "Drain" or something. That's the word everyone keeps throwing around when the feature is talked about. But maybe it makes complete sense if you put the amount into the input when max is clicked.

benkolera (approved):
LGTM. Just some nitpicks.

frontend/src/Frontend/UI/Dialogs/Send.hs
alexfmpe: Ah didn't really try, just moved this in from outside

benkolera: Ahh, cool.

alexfmpe: `_Wrapped` was acting up but `_Wrapped'` worked

frontend/src/Frontend/UI/Dialogs/Send.hs
alexfmpe: Looks like we're keeping this one with the same style as the read-only fields

PR #458, opened 2020-01-24
Make the back button do the right thing in the send dialog

benkolera:
This feels like it is missing something with remembering the uiMetaData state and repopulating it, but I must admit that that bit really confuses me because we ignore everything from the uiMetaData call except for the mConf.

Someone that knows that better, please pull me up if there is something missing here still. :)

### DIFF
--- a/frontend/src/Frontend/UI/Dialogs/Send.hs
+++ b/frontend/src/Frontend/UI/Dialogs/Send.hs
@@ -26,11 +26,13 @@ import Control.Lens hiding (failover)
 import Control.Monad (join, when, void, (<=<))
 import Control.Monad.Except (throwError)
 import Control.Monad.Logger (LogLevel(..))
+import Control.Monad.Trans (lift)
 import Control.Monad.Trans.Except
 import Control.Monad.Trans.Maybe
 import Data.Bifunctor (first)
 import Data.Decimal (Decimal)
-import Data.Dependent.Sum ((==>))
+import Data.Dependent.Sum ((==>),DSum((:=>)))
+import Data.Dependent.Map(Some(Some))
 import Data.Either (isLeft, rights)
 import Data.List.NonEmpty (NonEmpty(..))
 import Data.Text (Text)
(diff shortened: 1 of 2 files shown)


frontend/src/Frontend/UI/Dialogs/Send.hs
```diff
@@ -260,24 +268,57 @@ sameChainTransfer model netInfo keys fromAccount gasPayer toAccount amount = Wor
...
+            (toAccount, amount) <- MaybeT $ fmap hush $ sample $ current recipient
+            let mCrossChain = ffor mToGasPayer $ \a -> CrossChainData
+                  { _crossChainData_recipientChainGasPayer = a }
+            let transfer = TransferData
+                  { _transferData_amount = amount
+                  , _transferData_crossChainData = mCrossChain
+                  , _transferData_fromAccount = fromAccount
+                  , _transferData_fromGasPayer = fromGasPayer
+                  , _transferData_toAddress = toAccount
+                  }
```
mankyKitty, line 311: as `Dynamic` has a Monad instance too, you could probably avoid the explicit `sample $ current`s and have something like:
```haskell
else do
            (fromGasPayer, mToGasPayer) <- lift mCaps
            (toAccount, amount) <- lift $ fmap hush recipient
            let mCrossChain = ffor mToGasPayer $ \a -> CrossChainData
                  { _crossChainData_recipientChainGasPayer = a }
            let transfer = TransferData
                  { _transferData_amount = amount
                  , _transferData_crossChainData = mCrossChain
                  , _transferData_fromAccount = fromAccount
                  , _transferData_fromGasPayer = fromGasPayer
                  , _transferData_toAddress = toAccount
                  }
```
??

benkolera: I've heard that a puppy dies every time you use the Dynamic monad instance tho... 

The sample current stuff is not new code though, so this could be intentional. @tomsmalley ?

benkolera:
Closing this PR because it's against develop and not 2020 and there's some awful conflicts. I'll just port this across and make a new PR.

frontend/src/Frontend/UI/Dialogs/Send.hs
tomsmalley: No need to use the Dynamic instance, you can use the Behavior instance with `tagMaybe`. They end up doing the same thing in the end, but the tag version is a little cleaner I suppose.

PR #461, opened 2020-01-24
Makes going back from the send preview remember form state and go back to the sign tab

benkolera:
Supercedes https://github.com/kadena-io/chainweaver/pull/458


frontend/src/Frontend/UI/Dialogs/Send.hs
```diff
@@ -252,25 +274,40 @@ sameChainTransfer model netInfo keys (fromName, fromChain, fromAcc) (gasPayer, g
...
-        (fromGasPayer, mToGasPayer) <- MaybeT $ sample $ current mCaps
-        (toAccount, amount) <- MaybeT $ fmap hush $ sample $ current recipient
-        let mCCD = case mToGasPayer of
-              Just toGasPayer | toGasPayer /= fromGasPayer -> Just $ CrossChainData
-                { _crossChainData_recipientChainGasPayer = toGasPayer
-                }
-              _ -> Nothing
-        pure $ previewTransfer model fromAccount fromGasPayer toAccount mCCD amount
+        currTab <- lift $ sample $ current currentTab
+        if (currTab /= SendModalTab_Sign)
```
alexfmpe: Since we end with 
```haskell
pure $ previewTransfer model transfer
```
how about having `footerSection` return `Event t SendModalTab` instead of `Event t ()` and make this be
```haskell
let nextScreen = flip push next $ \nextTab -> runMaybeT $ do
  if (nextTab /= SendModalTab_Preview)
```

That way the decision of which is the next screen based on the current one is in only one place.
Right now we do two separate checks: one for the 'next' button and another for the modal body.

benkolera: Done (well, preview isn't a tab, so it became `Event t (Maybe SendModalTab)` but that's nbd. Also, conflicts resolved.

tomsmalley:
LGTM

PR #462, opened 2020-01-24
Make the back button from record recovery phrase go to the passphrase warning

benkolera:
Reported by QA here: https://app.asana.com/0/1151408343688145/1158353567309625/f 

Happy to have this argued as not a bug and it is intended to go back to the password setup. When testing it, I'm not sure that it actually makes sense to go back to the passphrase remembrance warning. If the user is going back it's because they either want to set the password different or do a restore instead of create.


tomsmalley:
Closing as not a bug

PR #467, opened 2020-01-27
Apply 'Event in, Dynamic out'

mankyKitty:


frontend/src/Frontend/UI/Dialogs/AddVanityAccount/DefineKeyset.hs
```diff
@@ -32,8 +32,8 @@ data KeysetInputs t i a = KeysetInputs
   }
 
 data DefinedKeyset t = DefinedKeyset
-  { _definedKeyset_internalKeys :: KeysetInputs t (Dropdown t (Maybe Int)) PublicKey
-  , _definedKeyset_externalKeys :: KeysetInputs t (ExternalKeyInput t) PublicKey
+  { _definedKeyset_internalKeys :: KeysetInputs t (Maybe (Maybe Int)) PublicKey
```
tomsmalley: I feel like the stacked maybe is worth a comment explaining what the difference between `Nothing` and `Just Nothing` is

frontend/src/Frontend/UI/Dialogs/AddVanityAccount/DefineKeyset.hs
```diff
@@ -32,8 +32,8 @@ data KeysetInputs t i a = KeysetInputs
...
+  , _definedKeyset_externalKeys :: KeysetInputs t (Maybe (Maybe PublicKey)) PublicKey
```
tomsmalley: Likewise here

PR #469, opened 2020-01-28
Rename 'Kadena Address' to 'Tx Builder'

mankyKitty:


frontend/src/Frontend/UI/Dialogs/AddVanityAccount.hs
```diff
@@ -192,13 +192,13 @@ createAccountNotGasPayer
 createAccountNotGasPayer ideL name chain selectedKeyset keyset = Workflow $ do
   modalMain $ do
     dialogSectionHeading mempty "Notice"
-    divClass "group" $ text "The text below contains all of the Account info you have just configured. Share this [Kadena Address] with someone else to pay the gas for the transaction to create the Account."
```
tomsmalley, line 195: I wonder if the square brackets were to denote a placeholder name?

tomsmalley: Confirmed by Taylor, square brackets should be removed now

mankyKitty: Fixed.

PR #471, opened 2020-01-28
Use git hash as mac app bundle version

tomsmalley:
Technically this isn't allowed, but it doesn't seem to cause any
issues...
https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364


tomsmalley:
Closing this in favour of adding the version in the UI somewhere consistent between platforms.

Also, the method in this PR means versions don't appear in the Finder.

PR #480, opened 2020-01-29
Remove unused argument

tomsmalley:
Well, hide the warning.


alexfmpe:
Would it be better to instead not call this function instead of commenting it all the way down into `pure ()`?

tomsmalley:
I don't disagree, but it seems likely that we end up re-adding the function at some point

PR #484, opened 2020-01-29
Consistent versioning across all apps

benkolera:
Mac means that we need a X.Y.Z version. 

This makes the "chainweaver" version the X.Y part that is shared across all apps, but lets us make platform specific changes in the Z number. 

X is reserved for big UI changes, where Y is for smaller features and non platform specific fixes.

All in the toplevel default.nix so that all versions are changed in those handful of lines so things aren't forgotten.

We're dropping the "Beta" prefix and going to 1.0 as the first real proper version. A little odd because the last mac version was 1.5.1.2 but no one will mind. Anagha and Doug are on board with this being the next version number.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -8,8 +8,8 @@ with obelisk;
 let
   # All the versions that the user cares about are here so that they can
   # be changed in one place
-  chainWeaverVersion = "1.0";
-  appName = "Kadena Chainweaver ${chainWeaverVersion}";
+  chainweaverVersion = "1.0";
+  appName = "Kadena Chainweaver ${chainweaverVersion}";
   macReleaseNumber = "0";
   linuxReleaseNumber = "0";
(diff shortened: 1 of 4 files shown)


default.nix
```diff
@@ -6,21 +6,27 @@ args@{ system ? builtins.currentSystem
 }:
 with obelisk;
 let
+  # All the versions that the user cares about are here so that they can
+  # be changed in one place
+  chainWeaverVersion = "1.0";
```
alexfmpe: Huh never saw the `w` in `Chainweaver` capitalized

benkolera: Yeah, good point. I'll update. I didn't do this out of any reason. Just the first thing I did and autocomplete did the rest, haha.

alexfmpe (approved):

PR #495, opened 2020-01-31
Workaround missing row bug

tomsmalley:


benkolera:
I am deeply disturbed and puzzled as to why that slight refactoring works around the issue, haha.

PR #498, opened 2020-02-03
Make user choose keys rather than accounts whenever signing generic contract deployments

alexfmpe:
On top of #492 & #497


frontend/src/Frontend/UI/DeploymentSettings.hs
```diff
@@ -294,30 +274,21 @@ buildDeploymentSettingsResult m mSender mGasPayer signers cChainId capabilities
...
-    let signingPairs = getSigningPairs chainId keys allAccounts signingAccounts
     cmd <- buildCmd
       (_deploymentSettingsConfig_nonce settings)
-      networkId publicMeta signingPairs
+      networkId publicMeta signingKeypairs
       (_deploymentSettingsConfig_extraSigners settings)
-      code' (HM.union jsonData' deploySettingsJsonData) pkCaps
-    wrappedCmd <- for (wrapWithBalanceChecks signingAccounts code') $ \wrappedCode -> do
+      code' (HM.union jsonData' deploySettingsJsonData) publicKeyCapabilities
+    for_ (wrapWithBalanceChecks (Set.singleton sender) code') $ \wrappedCode -> do
```
tomsmalley: No point doing this just to throw it away

PR #504, opened 2020-02-03
Filters gas payers to accounts that have a nonzero balance on that chain

benkolera:
This explicitly doesn't change anything for the signing / deployment sender dropdowns. 

They feel like they need some work in this brave new accounts 2020 world, but this is not the PR to tidy them up. :)


frontend/src/Frontend/UI/Widgets.hs
```diff
@@ -913,19 +914,38 @@ uiAccountFixed sender = do
...
+        | Just a <- Map.lookup chain chainMap
+        , not needsGas || (fromMaybe 0 (a ^? account_status . _AccountStatus_Exists . accountDetails_balance)) > 0
+        = Map.singleton n (unAccountName n)
         | otherwise = mempty
       vanityAccounts = Map.foldMapWithKey mkVanity accountsOnNetwork
       accountsOnChain = vanityAccounts
   when (Map.null accountsOnChain) $ throwError "No accounts on current chain"
   pure accountsOnChain
+  -- where
+  --  calcMaxGas :: PublicMeta -> Decimal
```
alexfmpe: Leftovers

frontend/src/Frontend/UI/Widgets.hs
```diff
@@ -913,19 +914,38 @@ uiAccountFixed sender = do
...
   netId <- lift $ m ^. network_selectedNetwork
+  -- Calculating the max gas from the network meta causes a causality loop
+  -- maxGasPriceBalance <- lift $ AccountBalance . calcMaxGas <$> m ^. network_meta
+
   chain <- ExceptT $ note "You must select a chain ID before choosing an account" <$> mChainId
   accountsOnNetwork <- ExceptT $ note "No accounts on current network" . Map.lookup netId . unAccountData <$> m ^. wallet_accounts
   let mkVanity n (AccountInfo _ chainMap)
-        | Map.member chain chainMap = Map.singleton n (unAccountName n)
+        | Just a <- Map.lookup chain chainMap
+        , not needsGas || (fromMaybe 0 (a ^? account_status . _AccountStatus_Exists . accountDetails_balance)) > 0
```
alexfmpe: This will now fail if the account isn't found, though that sounds more or less impossible?

alexfmpe: Actually seems to be an improvement, given the error message we show.

alexfmpe (approved):

PR #507, opened 2020-02-04
Apply popover feedback mechanism

mankyKitty:
Replaced most of the existing error/feedback mechanisms with the new popover style.

Need feedback on table CSS and on the workings of the capabilities input row feedback.

### DIFF
--- a/backend/sass/common.blocks/_modal.scss
+++ b/backend/sass/common.blocks/_modal.scss
@@ -194,9 +194,7 @@ body[data-modal-open='true'] {
   }
   .table td {
     max-width: 0;
-    // overflow: hidden;
     text-overflow: ellipsis;
-    // white-space: nowrap;
   }
 }
 
--- a/backend/sass/common.blocks/_modal.scss
+++ b/backend/sass/common.blocks/_modal.scss
@@ -194,9 +194,9 @@ body[data-modal-open='true'] {
   }
   .table td {
     max-width: 0;
-    overflow: hidden;
+    // overflow: hidden;
     text-overflow: ellipsis;
-    white-space: nowrap;
+    // white-space: nowrap;
   }
 }
 
--- a/frontend/src/Frontend/UI/Dialogs/Send.hs
+++ b/frontend/src/Frontend/UI/Dialogs/Send.hs
@@ -77,7 +77,7 @@ import Frontend.UI.Dialogs.DeployConfirmation (submitTransactionWithFeedback)
 import Frontend.UI.Modal
 import Frontend.UI.TabBar
 import Frontend.UI.Widgets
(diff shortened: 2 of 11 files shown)


frontend/src/Frontend/UI/Widgets.hs
```diff
@@ -1111,8 +1099,8 @@ uiInputWithPopover body getStateBorderTarget mkMsg cfg = divClass "popover" $ do
 
   _ <- runWithReplace (divClass "popover__message" blank) $ leftmost
     [ popoverDiv . popoverToAttrs <$> onMsg
-    , popoverDiv popoverHiddenAttrs <$ onBlur
-    , popoverDiv . popoverToAttrs <$> current dPopState <@ onFocus
+    -- , popoverDiv popoverHiddenAttrs <$ onBlur
```
tomsmalley: Leftovers?

backend/sass/common.blocks/_modal.scss
```diff
@@ -194,9 +194,9 @@ body[data-modal-open='true'] {
   }
   .table td {
     max-width: 0;
-    overflow: hidden;
+    // overflow: hidden;
     text-overflow: ellipsis;
-    white-space: nowrap;
+    // white-space: nowrap;
```
tomsmalley: Leftover?

frontend/src/Frontend/UI/Widgets.hs
mankyKitty: They are required, I'll uncomment them. That's a debugging leftover because otherwise it's really difficult to debug the styling.

frontend/src/Frontend/UI/Dialogs/Send.hs
```diff
@@ -317,33 +317,54 @@ sendConfig model initData = Workflow $ do
...
+                      . accountDetails_balance
+                      . _AccountBalance
 
-          (_, amount, _) <- mkLabeledInput True "Amount" uiGasPriceInputField
-            (def & inputElementConfig_initialValue .~ (maybe "" tshow mInitAmount))
-            -- We're only interested in the decimal of the gas price
-            <&> over (_2 . mapped . mapped) (\(GasPrice (ParsedDecimal d)) -> d)
+                    (GasPrice (ParsedDecimal gp)) = gp0
+                in
+                  if maybe True (gp >) fromBal then
```
alexfmpe, line 361: I'm not sure we should claim insufficient funds if the balance is unknown.

alexfmpe: Turns out the current spec means the balance is always known to exist, otherwise we wouldn't be able to be in this dialog. Fixing that in another PR

PR #511, opened 2020-02-05
Rework frontend/backend package/bindings organization

alexfmpe:


backend/src/Backend.hs
```diff
@@ -56,13 +63,24 @@ import           Common.OAuth              (OAuthProvider (..),
...
 
+main :: IO ()
+main = do
+  runCheck <- not . null . filter (== "check-deployment") <$> getArgs
+  if runCheck
+     then checkDeployment
+     else runBackend backend webFrontend
+
+frontend :: Frontend (R FrontendRoute)
+frontend = webFrontend
```
alexfmpe: The intent was that we'd pick here (where we do not need to build under ghcjs, and thus can also have the `desktopFrontend` in scope) between multiple frontends, rather than having dedicated infra in obelisk for that, but that last step is currently under discussion.

.obelisk/impl/github.json
```diff
@@ -2,6 +2,6 @@
   "owner": "obsidiansystems",
   "repo": "obelisk",
   "branch": "chainweaver",
-  "rev": "d7fdaebad0e850249c84ceb4b63613d4dcc0a692",
-  "sha256": "1h8mrhccxxjqvnbdq25raf6mhigbzrv9glpg5c8sdqkh4anl0ap8"
+  "rev": "aead55c71d76d44e9e6ad2b28834b61e10974448",
```
alexfmpe: https://github.com/obsidiansystems/obelisk/compare/d7fdaebad0e850249c84ceb4b63613d4dcc0a692...aead55c71d76d44e9e6ad2b28834b61e10974448

.obelisk/impl/github.json
```diff
@@ -2,6 +2,6 @@
   "owner": "obsidiansystems",
   "repo": "obelisk",
   "branch": "chainweaver",
-  "rev": "d7fdaebad0e850249c84ceb4b63613d4dcc0a692",
-  "sha256": "1h8mrhccxxjqvnbdq25raf6mhigbzrv9glpg5c8sdqkh4anl0ap8"
+  "rev": "00965d65e0557110efa34a371e8dc0c59c9f29d1",
```
alexfmpe, line 5: https://github.com/obsidiansystems/obelisk/compare/d7fdaebad0e850249c84ceb4b63613d4dcc0a692...00965d65e0557110efa34a371e8dc0c59c9f29d1

PR #526, opened 2020-02-06
Blk@chainweaver 1.0.0 rc/ova 1909

benkolera:
Brings in the nixos 19.09 fixes in so the version script is the name that was preferred and the autostart is future proof.


benkolera:
On second thoughts, with the s3 pointer being OK in the release, lets not change stuff at the last minute.

PR #527, opened 2020-02-06
Allow user to change password

tomsmalley:


desktop/src/Desktop/Crypto/BIP.hs
```diff
@@ -57,12 +58,12 @@ bipCryptoGenPair root pass i =
     scheme = Crypto.DerivationScheme2
     mkHardened = (0x80000000 .|.)
 
-instance MonadJSM m => HasCrypto Crypto.XPrv (BIPCryptoT m) where
+instance (MonadSample t m, MonadJSM m) => HasCrypto Crypto.XPrv (BIPCryptoT t m) where
   cryptoSign bs xprv = BIPCryptoT $ do
-    (_root, pass) <- ask
+    (_, pass) <- sample =<< ask
```
tomsmalley, line 63: This is the only thing I'm not particularly happy about. It works fine though.

PR #534, opened 2020-02-07
Disable unrestricted signers which are in caps list

tomsmalley:


frontend/src/Frontend/UI/DeploymentSettings.hs
```diff
@@ -727,19 +728,31 @@ uiSignerList
...
+      pb <- getPostBuild
+      let conf = def
+            & checkboxConfig_attributes .~ ffor preSelected (\s -> if s then "disabled" =: "disabled" else mempty)
+            & checkboxConfig_setValue .~ leftmost
+              [ updated preSelected
+              , tag (current preSelected) pb
+              ]
+      cb <- uiCheckbox "signing-ui-signers__signer" False conf $
+        text $ keyToText $ _keyPair_publicKey pair
+      pure $ xor <$> _checkbox_value cb <*> preSelected
```
alexfmpe: Let's clarify whether "checked off" means "checked on" or "checked = off"

PR #540, opened 2020-02-11
Add dialog for watching requests

tomsmalley:


frontend/src/Frontend/UI/Dialogs/WatchRequest.hs
```diff
@@ -0,0 +1,116 @@
...
+  choice <- modalMain $ do
+    dialogSectionHeading mempty "Notice"
+    divClass "group" $ do
+      text "If you have a request key from a previously submitted transaction, you can use this dialog to wait for and display the results."
+    dialogSectionHeading mempty "Required Information"
+    info <- divClass "group" $ do
+      mChain <- mkLabeledClsInput False "Chain ID" $ uiChainSelection (headMay <$> nodes) (pure Nothing)
+      eKey <- mkLabeledInput False "Request Key" uiInputElement def
+      let parseRequestKey t
+            | {- T.length t == 43, -} Right v <- Pact.fromText' t = Right v
```
alexfmpe: eh?

tomsmalley: All the hashes appear to be 43 characters long. It seemed weird using the dialog without this, because you could pretty much enter anything you wanted and progress to the next screen. I'm not sure if these hashes will always be that long, though, or if the few times I did transactions they just happened to be that long. Removed the comment anyway, I forgot I'd left it in.

PR #546, opened 2020-02-12
Import / Export Functionality

benkolera:
So this is ready to go. Any fixes are in @mankyKitty 's hands now, but I'm around if there are questions! 

The biggest todo is that the transaction log entry for import / export isn't being written yet. That's something that should be done in the branch with the txn log. 

The bug on devDevelop where if you choose the same file twice nothing happens on the second time... that's been around forever so not worth fixing here. 

This also sneakily adds a password change success screen to change password so Tom and my dialogs are aligned.

TODO:
- [x] Styling of the Import Form
- [x] Enter key on export
- [x] Check password on import
- [x] Mac File Chooser / Deliverer
- [x] Linux File Chooser / Deliverer
- [ ] Log entry for export import
- [x] Success pages for pw change dialog

Bugs:
- [x] Make the key in the filename just the first 8 chars
- [ ] You can only pick one import once without refreshing (try restoring from import twice without refreshing. A file cannot be selected. It only affects devDesktop though). This bug existed before for the pact file chooser.
- [x] Bad alignment of folder icon of export on settings page


mankyKitty:
superseded by #562

PR #573, opened 2020-02-25
Rudimentary handling of accounts with non-keyset guards

mankyKitty:
Adds support for non-keyset address keysets. Only tacks it on the end at the moment to see
if that fixes the current problem, whilst thinking about better solutions

### DIFF
--- a/frontend/src/Frontend/Network.hs
+++ b/frontend/src/Frontend/Network.hs
@@ -73,7 +73,6 @@ module Frontend.Network
   , HasTransactionLogger(..)
   ) where
 
-import Debug.Trace
 import           Control.Exception                 (fromException)
 import           Control.Arrow                     (first, left, second, (&&&))
 import           Control.Lens                      hiding ((.=))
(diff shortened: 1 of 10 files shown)


frontend/src/Frontend/Network.hs
```diff
@@ -799,7 +800,7 @@ mkClientEnvs nodeInfos chain = fforMaybe nodeInfos $ \nodeInfo ->
 doReqFailover :: MonadJSM m => [S.ClientEnv] -> S.ClientM a -> m (Either [S.ClientError] a)
 doReqFailover [] _ = pure $ Left []
 doReqFailover (c:cs) request = liftJSM $ S.runClientM request c >>= \case
-  Left e -> BiF.first (e:) <$> doReqFailover cs request
+  Left e -> BiF.first (traceShowId e:) <$> doReqFailover cs request
```
alexfmpe: Leftover

frontend/src/Frontend/UI/Dialogs/Send.hs
```diff
@@ -540,7 +540,7 @@ runUnfinishedCrossChainTransfer logL netInfo keys fromChain toChain toGasPayer r
   let (continueError, continueOk) = fanEither continueResponse
   -- Wait for result
   resultResponse <- listenForSuccess logL envToChain continueOk
-  let (resultError, resultOk) = fanEither resultResponse
+  let (resultError, resultOk) = fanEither $ traceEvent "resultResponse" resultResponse
```
alexfmpe: Leftover

PR #581, opened 2020-02-28
WIP: Handle the possibility of existing same-version log files

mankyKitty:
If there exists a log file of the name we want to write to then find how many we're up to
and move the most recent one out of the way so we can freely write the imported log file.


mankyKitty:
Closing, redundant from other changes.

PR #587, opened 2020-03-02
Add path to chainweaver service so link opening works

mankyKitty:


benkolera (approved):
Definitely a hack, lol, but fixing this is priority over the nicer solutions.

There is almost certainly some stuff in home manager for making sure the user path is right for user services because it even sets up emacs servers and what not.

I think there was a reason why I didn't use the home manager way, but I can't remember. Probably just time pressure and pebcak.

PR #589, opened 2020-03-03
Fix up spacing on external key inputs

mankyKitty:


frontend/src/Frontend/UI/Dialogs/AddVanityAccount/DefineKeyset.hs
```diff
@@ -61,23 +61,25 @@ uiExternalKeyInput onPreselection = do
     uiPubkeyInput iv = do
       let
         inp cfg = do
-          ie <- uiInputElement cfg
+          ie <- mkLabeledInput False  mempty uiInputElement  cfg
```
alexfmpe: Ironic extra whitespace

mankyKitty: :man_facepalming:

PR #592, opened 2020-03-03
Manual TxBuilder

mankyKitty:


frontend/src/Frontend/UI/Widgets.hs
```diff
@@ -878,29 +883,39 @@ uiGasPriceInputField eReset conf = dimensionalInputFeedbackWrapper (Just "KDA")
...
+    mNodeInfo = (^? to rights . _head) <$> m ^. network_selectedNodes
 
 uiChainSelection
   :: MonadWidget t m
   => Dynamic t (Maybe NodeInfo)
   -> Dynamic t (Maybe ChainId)
   -> CssClass
-  -> m (Dynamic t (Maybe ChainId))
+  -> m ( Dropdown t (Maybe ChainId)
+       , Dynamic t (Maybe ChainId)
```
alexfmpe: Unlike when we return `(Event t, Dynamic t)`, these are redundant, but callers don't know that. I'd just return `Dropdown t`.

PR #601, opened 2020-03-09
Hardcode the GasLimit for the 'list-modules' function call

mankyKitty:


frontend/src/Frontend/Network.hs
```diff
@@ -645,7 +645,10 @@ loadModules logL networkL onRefresh = do
             (o:_, n:_) -> if o /= n then Just newInfos else Nothing
             _          -> Just newInfos
 
-      mkReq netName pm = mkSimpleReadReq "(list-modules)" netName pm . ChainRef Nothing
+      -- We need a minimum GasLimit to ensure that this call cannot fail. As the
+      -- PublicMeta stores the users last transaction configuration and may not be
+      -- sufficient for this function.
+      mkReq netName pm = mkSimpleReadReq "(list-modules)" netName (pm & Pact.pmGasLimit .~ 200) . ChainRef Nothing
```
alexfmpe, line 651: Can't we get this limit from a dependency? https://github.com/kadena-io/pact/blob/cf220977a6b50a47cd446dad2e5e444f61689672/src/Pact/Gas/Table.hs#L168 seems related

PR #602, opened 2020-03-10
Batch of polish items

mankyKitty:


frontend/src/Frontend/UI/Dialogs/Receive.hs
```diff
@@ -104,14 +104,15 @@ uiReceiveModal
...
+  => Maybe Text
+  -> model
   -> AccountName
   -> AccountDetails
   -> Maybe ChainId
   -> Event t ()
   -> m (mConf, Event t ())
-uiReceiveModal model account details mchain _onClose = do
-  onClose <- modalHeader $ text "Receive"
+uiReceiveModal titleOverride model account details mchain _onClose = do
```
alexfmpe: I'd just require `Text` everytime without a default. I mean, this could as well be called `uiTransferToModal` and "Receive" seen as a special case of "Transfer to"

PR #604, opened 2020-03-10
Add chain Id to preview screen for send preview

mankyKitty:


mankyKitty:
Still waiting on design approval, but marking as not a draft so others may merge when it is time.

PR #606, opened 2020-03-11
Refactor 'Watch Request' to 'Check TX Status'

mankyKitty:
Now uses `/poll` instead of `/listen` and does not require a ChainID.

### DIFF
--- a/frontend/src/Frontend/UI/Dialogs/WatchRequest.hs
+++ b/frontend/src/Frontend/UI/Dialogs/WatchRequest.hs
@@ -97,9 +97,9 @@ inputRequestKey model _ = Workflow $ do
 
       onTimeoutFinished <- delay 5 onRequest
 
-      blockSpam <- toggle False $ leftmost
-        [ onRequest
-        , onTimeoutFinished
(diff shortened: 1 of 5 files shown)


frontend/src/Frontend/UI/Dialogs/WatchRequest.hs
```diff
@@ -59,59 +68,86 @@ inputRequestKey
...
+
+    rec
+      let lockRefresh = (||) <$> fmap isNothing dRqKey <*> blockSpam
+
+      onRequest <- divClass "group check-tx-status__button-wrapper" $
+        confirmButton (def & uiButtonCfg_disabled .~ lockRefresh) checkTxBtnLbl
+
+      onTimeoutFinished <- delay 5 onRequest
+
+      blockSpam <- toggle False $ leftmost
```
alexfmpe: This would be more accurate as
```haskell
blockSpam <- holdDyn False $ leftmost
  [ True <$ onRequest
  , False <$ onTimeoutFinished
  ]
```
It seems equivalent if we never have more than one request in transit, but I don't think we want to add implicit assumptions on that.

PR #609, opened 2020-03-12
Make the icon clickable for settings items

mankyKitty:


alexfmpe:
The `div` that contains the icon and button shows the hand cursor but isn't clickable - it is in the resources screen.

PR #613, opened 2020-03-20
Auto-select chain when there's only one

alexfmpe:


frontend/src/Frontend/Network.hs
```diff
@@ -550,7 +551,12 @@ getNetworks cfg = do
          then Nothing
          else fst <$> Map.lookupMin networks
 
-
+-- Assumes all nodes have exactly the same chains
+-- TODO: Make the network dialog enforce that assumption, pull chains out of NodeInfo and remove this method
```
mankyKitty, line 555: Does this complete this TODO?

frontend/src/Frontend/UI/Widgets.hs
```diff
@@ -889,92 +886,49 @@ uiGasPriceInputField eReset conf = dimensionalInputFeedbackWrapper (Just "KDA")
...
   let
-    chains = map (id &&& _chainId) . maybe [] getChains <$> info
+    chains = map (id &&& _chainId) <$> options
     mkPlaceHolder cChains = if null cChains then "No chains available" else "Select chain"
     mkOptions cs = Map.fromList $ (Nothing, mkPlaceHolder cs) : map (first Just) cs
 
     staticCls = cls <> "select"
     mkDynCls v = if isNothing v then "select_mandatory_missing" else mempty
+  pb <- getPostBuild
+  let optionsEv = leftmost [current options <@ pb, updated options]
```
mankyKitty, line 910: Is this `tagOnPostBuild` ? It uses the `updated` event too, I think

alexfmpe: Oh, I hadn't realized `tagOnPostBuild` tagged _beyond_ on postBuild. But I see it uses `tagPromptlyDyn` so I'd rather steer clear of that.

PR #621, opened 2020-04-13
Query node health only on url edits, not on selections nor contract r…

alexfmpe:
…eloads


alexfmpe:
Subsumed by https://github.com/kadena-io/chainweaver/pull/628

PR #637, opened 2020-09-01
Fix apending of .0 to numbers

mightybyte:


frontend/src/Frontend/Foundation.hs
```diff
@@ -93,3 +95,14 @@ type family ReflexValue (f :: * -> *) x where
...
+addDecimalToString :: Text -> Text
+addDecimalToString s =
+    case T.find f s of
+      Nothing -> s <> ".0"
+      Just _ -> s
+  where
+    f '.' = True
+    f 'e' = True
+    -- Need to also check for 'e' because we don't want "1e-7" to be converted
+    -- to "1e-7.0"
```
emilypi: This looks okay, but there's a [parser for this in the `scientific` library](https://hackage.haskell.org/package/scientific-0.3.6.2/docs/Data-Scientific.html) which does it in a really clean way which could save us the trouble of writing the logic ourselves: 

```haskell
> import Text.ParserCombinators.ReadP (readP_to_S)
> import Data.Scientific (scientificP)
> readP_to_S scientificP "3"
[(3.0,"")]
> readP_to_S scientificP "3.0e2"
[(3.0,"e2"),(300.0,"")]
> readP_to_S scientificP "+3.0e+2"
[(3.0,"e+2"),(300.0,"")]
> readP_to_S scientificP "-3.0e-2"
[(-3.0,"e-2"),(-3.0e-2,"")]
```

which would amount to 

```haskell
appendDecimaltoText t = case readP_to_S scientificP t of 
    [] -> T.pack t
    [(x,_)] -> sshow x
    [_,(x,_)] -> sshow x
    _ -> error "impossible"
```

Which gets you 

```haskell
П> appendDecimaltoText "3"
"3.0"
П> appendDecimaltoText "3e-7"
"3.0e-7"
П> appendDecimaltoText "3e-2"
"3.0e-2"
П> appendDecimaltoText "3e+2"
"300.0"
П> appendDecimalToText "a"
"a"
```

mightybyte: Nice idea, I put that in.

frontend/src/Frontend/Foundation.hs
```diff
@@ -96,13 +98,11 @@ type family ReflexValue (f :: * -> *) x where
...
-addDecimalToString s =
-    case T.find f s of
-      Nothing -> s <> ".0"
-      Just _ -> s
+appendDecimalToText :: Text -> Text
+appendDecimalToText t = case readP_to_S scientificP s of 
+    [] -> T.pack s
+    [(x,_)] -> tshow x
+    [_,(x,_)] -> tshow x
+    _ -> error "impossible"
```
emilypi, line 106: Slight amendment: we can get rid of the impossible case altogether.

```haskell
appendDecimalToText :: Text -> Text
appendDecimalToText t = case readP_to_S scientificP s of 
    [(x,_)] -> tshow x
    [_,(x,_)] -> tshow x
    _ ->  T.unpack s
```

frontend/src/Frontend/Foundation.hs
```diff
@@ -96,13 +98,11 @@ type family ReflexValue (f :: * -> *) x where
...
-addDecimalToString :: Text -> Text
-addDecimalToString s =
-    case T.find f s of
-      Nothing -> s <> ".0"
-      Just _ -> s
+appendDecimalToText :: Text -> Text
+appendDecimalToText t = case readP_to_S scientificP s of 
+    [] -> T.pack s
+    [(x,_)] -> tshow x
+    [_,(x,_)] -> tshow x
```
emilypi, line 104: This case is not quite 1-1 with what you had. It will convert `"3e-7"` into `"3.0e-7"`. If we don't want that, then we can squash the `[]` case together with this one, and drop the error message: 


```haskell
appendDecimalToText :: Text -> Text
appendDecimalToText t = case readP_to_S scientificP s of 
    [(x,_)] -> tshow x
    _ -> T.unpack t
```

emilypi: We also don't need to impossible case in any example: 

```haskell
appendDecimalToText :: Text -> Text
appendDecimalToText t = case readP_to_S scientificP s of 
    [(x,_)] -> tshow x
    [_,(x,_)] -> tshow x
    _ -> T.unpack t
```

mightybyte: Fixed

emilypi (approved):

PR #644, opened 2021-04-28
Keyset-refs in Transfers

emmanueldenloye:
Enable transfers from accounts with keyset-refs in their guards


frontend/src/Frontend/TxBuilder.hs
```diff
@@ -5,22 +5,24 @@
...
-import           Pact.Types.Term (KeySet (..))
+import           Pact.Types.Term (KeySet (..), KeySetName(..))
 
 data TxBuilder = TxBuilder
   { _txBuilder_accountName :: AccountName
     -- ^ The account name associated with this address
   , _txBuilder_chainId :: ChainId
     -- ^ The chain where this account resides
-  , _txBuilder_keyset :: Maybe KeySet
+  , _txBuilder_keyset :: Maybe (Either KeySetName KeySet)
```
emmanueldenloye: Probably need to rename this field name.

common/src/Common/Wallet.hs
```diff
@@ -241,6 +242,8 @@ data AccountStatus a
 data AccountGuard
   = AccountGuard_KeySet (Set PublicKey) Text
   -- ^ Keyset guards
+  | AccountGuard_KeySetRef Text
```
mightybyte: This new constructor seems problematic because it creates incomplete pattern matches in other parts of the code.  Maybe now is a good time to think about the comment above about replacing this with pact's representation.

Although...maybe this is a good opportunity for a middle ground.  Perhaps this AccountGuard data type can become the abstraction we're needing now that abstracts over the difference between KeySet and KeySetRef.

```
data KeySetHeritage = KeySetHeritage
  { keySetHeritage_keys :: Set PublicKey
  , keySetHeritage_pred :: Text
  , keySetHeritage_ref :: Maybe Text
  }

data AccountGuard
  = AccountGuard_KeySetLike KeySetHeritage
  | AccountGuard_Other (Pact.Guard PactValue)
```

This abstraction would make it so that after you look up a keyset-ref's underlying keyset, any code that is downstream from that point will behave the same.  That's probably what we want here, otherwise we'll probably encounter this same bug again in the future.

frontend/src/Frontend/UI/Transfer.hs
```diff
@@ -461,9 +463,57 @@ lookupKeySets logL networkName nodes chain accounts = do
...
+  -- TODO: CLEAN THIS UP!
+    rr <- case r of
+      Just res -> fmap Just $ iforM res $ \(AccountName name) dets -> do
+        let ref' = dets ^? _AccountStatus_Exists . accountDetails_guard . _AccountGuard_KeySetRef
+        let bal' = dets ^? _AccountStatus_Exists . accountDetails_balance . (to unAccountBalance)
+            updateBal old new = if old == new then old else new
+        case (,) <$> bal' <*> ref' of
+          Just (bal, ref) -> do
+            let describeCode keysetref account =
+                  [NeatInterpolation.text|{"balance" : (at 'balance (coin.details "$account")), "guard" : (describe-keyset "$keysetref")}|]
```
mightybyte: This is not a very big string literal, so let's just use `printf` and avoid the dependency on `neat-interpolation`.

frontend/src/Frontend/UI/Transfer.hs
```diff
@@ -1280,6 +1330,8 @@ getKeysetActionSingle c (Just a) mDetails =
     Just (AccountStatus_Exists (AccountDetails _ g)) -> do
       case g of
         AccountGuard_Other _ -> KeysetNoAction
+        -- TODO: needs better comment below lol
+        AccountGuard_KeySetRef refName -> KeysetNoAction -- this should be resolved at an earlier step
```
mightybyte: Here is a great example of where the above `KeySetHeritage` type and `KeySetLike` constructor would be useful.  The comment here says that the keyset-ref should be resolved at an earlier step, but it's not immediately obvious from a quick review whether that is actually true or not.  It's possible that doing `KeysetNoAction` here covers up some important cases.  But if you had `KeySetLike ksh`, then all the needed information that should have been resolved earlier would definitely be included.

frontend/src/Frontend/UI/Wallet.hs
```diff
@@ -379,6 +379,8 @@ accountGuardSummary (AccountGuard_Other pactGuard) =
   where
     gType = pactGuardTypeText $ Pact.guardTypeOf pactGuard
 
+accountGuardSummary (AccountGuard_KeySetRef name) = "keyset-ref: " <> name
```
mightybyte: Let's just say "ref: " here.  A lot of times the widgets that call `accountGuardSummary` are space-constrained.

mightybyte (changes requested):

common/src/Common/Wallet.hs
emmanueldenloye: I agree. I've included this data definition in the code.

frontend/src/Frontend/UI/Transfer.hs
emmanueldenloye: I agree. Since we now use `KeySetHeritage`, we either hit the `AccountGuard_Other` case (which means that keyset-ref has not been resolved, which is possible, if the keyset-ref is invalid) or we hit the case `AccountGuard_KeySetLike` which includes the `KeySetHeritage` in its constructor, which could contain a keyset-ref along with its accompanying keyset.

frontend/src/Frontend/UI/Transfer.hs
emmanueldenloye: OK, using `printf` now.

common/src/Common/Wallet.hs
```diff
@@ -239,17 +243,26 @@ data AccountStatus a
 -- interested in keyset guards right now. Someday we might end up replacing this
 -- with pact's representation for guards directly.
 data AccountGuard
-  = AccountGuard_KeySet (Set PublicKey) Text
-  -- ^ Keyset guards
+  = AccountGuard_KeySetLike KeySetHeritage
+  -- ^ Keyset guards. This may contain the name of the ke
```
mightybyte: Looks like this comment got cut off.

common/src/Common/Wallet.hs
```diff
@@ -267,15 +280,21 @@ instance FromJSON AccountGuard where
       keySet = withObject "AccountGuard" $ \o -> do
         keys <- o .: "keys"
         ksPred <- o .: "pred"
-        case mkAccountGuard keys ksPred of
+        ksRef <- o .:? "ref"
```
mightybyte: We need to be very careful about these JSON serialization changes.  If they're persisted anywhere and not completely backwards compatible, that will cause problems.  Before merging this we need to understand all the places where this serialization is used and assess the potential impact.

scripts/dev
```diff
@@ -11,7 +11,7 @@
 is_darwin() { [ "$(uname -s)" == 'Darwin' ]; }
 
 if is_darwin; then
-  ob run --interpret desktop --interpret mac
+  ob run --import desktop:Desktop --import mac:Mac
```
mightybyte: It looks like there was a merge issue somewhere and this old code came back.

frontend/src/Frontend/UI/Transfer.hs
```diff
@@ -461,9 +460,50 @@ lookupKeySets logL networkName nodes chain accounts = do
...
+                    return res
+                  Pact.PactResult (Left e) -> do
+                    putLog logL LevelInfo $ "lookupKeySets failed on keyset-ref lookup:" <> tshow e
+                    pure details
+
+
+    liftIO $ trigger resolvedReq
   pure result
 
+-- no signing of any kind here
```
mightybyte: Let's move this into `Frontend.Network` so it will be easier for other people to discover later.

frontend/src/Frontend/UI/Transfer.hs
```diff
@@ -27,6 +27,7 @@ Design Requirements:
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE QuasiQuotes #-}
```
mightybyte: This shouldn't be needed any more.

frontend/src/Frontend/UI/Transfer.hs
```diff
@@ -87,9 +86,9 @@ import           Pact.Types.PactValue
 import           Pact.Types.Pretty
 import           Pact.Types.RPC
 import           Pact.Types.Scheme
-import           Pact.Types.Term (KeySet (..))
+-- import           Pact.Types.Term (KeySet (..))
```
mightybyte: Let's just ditch these instead of commenting.

mightybyte (changes requested):
Looks great!  This handles the keyset-ref case in a nice self-contained way.  Just a few more small comments below.

frontend/src/Frontend/UI/Dialogs/AddVanityAccount.hs
```diff
@@ -358,7 +182,14 @@ createAccountConfig ideL renderErrors splashWF name chainId keyset = Workflow $
       Nothing
       $ uiAccountDropdown def (pure $ \_ a -> fromMaybe False (accountHasFunds a)) (pure id)
 
-  let payload = HM.singleton tempkeyset $ toJSON keyset
+  let payload = HM.singleton tempkeyset $ toJSON' keyset
```
mightybyte: I think this whole function is unused now.

emmanueldenloye: good catch

frontend/src/Frontend/UI/Dialogs/AddVanityAccount.hs
```diff
@@ -45,6 +45,8 @@ import Frontend.Foundation
 import Frontend.Log
 import Frontend.TxBuilder
 
+import Pact.Types.ChainMeta (PublicMeta (..), TTLSeconds)
```
mightybyte: One more...is this still used?

frontend/src/Frontend/UI/Dialogs/Receive/Legacy.hs
```diff
@@ -6,7 +6,8 @@ module Frontend.UI.Dialogs.Receive.Legacy
   , HasNonBIP32TransferInfo (..)
   , uiReceiveFromLegacy
   , receiveFromLegacySubmitTransfer
-  , receiveFromLegacySubmitTransferCreate
+  , receiveFromLegacySubmit
+  -- , receiveFromLegacySubmitTransferCreate
```
mightybyte: And let's get rid of this too.

frontend/src/Frontend/UI/Dialogs/AddVanityAccount.hs
emmanueldenloye: No, I guess not.

mightybyte (approved):

PR #647, opened 2021-05-11
Chainweaver 2.1.1

emmanueldenloye:


jmininger:
Tested both OVA and .deb created in a local nix-build and they each have `api.chainweb.com` as the default nodes

mightybyte:
I tested the mac `.dmg` and it also defaults new wallets to `api.chainweb.com` as desired.

PR #648, opened 2021-05-12
Populate receive modal for accts with keyset-refs

emmanueldenloye:


frontend/src/Frontend/TxBuilder.hs
```diff
@@ -24,6 +24,7 @@ data TxBuilder = TxBuilder
     -- ^ Presence or absence of a keyset may be used to determine transfer vs
     -- transfer-create. If the keyset is present and the account already exists
     -- you could choose to do either a transfer or a transfer-create.
+  , _txBuilder_keysetRef :: Maybe KeySetName
```
mightybyte, line 27: The `TxBuilder` type is very sensitive because it defines the serialization format that is used both by Chainweaver and by other parts of the ecosystem.  We don't want keyset refs to be visible at this level, both for backwards compatibility reasons and because the low level details of keyset refs should be hidden from users of the TxBuilder.  So I think we should back out the changes to this data type and make sure that Chainweaver resolves the keyset elsewhere, preserving the keyset interface.

mightybyte (changes requested):

Issue #646, opened 2021-05-11
Cross-chain transfers sometimes don't finish


mightybyte:
Polling and  retry logic probably needs some work

PR #649, opened 2021-05-14
Switch cross-chain polling mechanism from listen to poll

jmininger:
Switch from a `listen` call to a manual `poll` call for both steps of the cross-chain transfer mechanism (Listening for continuation, listening for transfer). Chainweb's `defaultTimeout` is approximately 2 minutes, and this change sets the default polling behavior to 12 attempts spaced out every 15 seconds (until success). This effectively makes it a 3 minute polling period. Addresses #646


frontend/src/Frontend/UI/Dialogs/Send.hs
```diff
@@ -1026,49 +1055,62 @@ listenForContinuation
...
-        | otherwise -> pure $ Left $ tshow $ peDoc pe
-      Pact.PactResult (Right _) -> pure $ Right ()
-    Right (Api.ListenTimeout _) -> pure $ Left "Listen timeout"
-  liftIO $ cb r
+  -> Event t (Maybe Pact.PactId, Pact.RequestKey)
+  -> Int --max polls
+  -> Int -- delay between polls (seconds)
+  -> (Pact.CommandResult Hash -> Maybe Pact.PactId -> Either Text a) -- callback for handling poll result
+  -> m (Event t (Either Text a))
+pollNodesForReq model envs requestPair maxPolls waitTime handler =
```
emmanueldenloye, line 1084: If you don't mind me suggesting, have you used the `retry` library before? It is really good for forming delay combinators for such situations such as these?

mightybyte: Yeah, I like the `retry` library as well but in this case I'm not sure it's worth it.  We don't need sophisticated delay strategies here and it's since it's a web app I'd probably say that incurring another dependency is worse than the relatively straightforward loop to implement this logic.

frontend/src/Frontend/UI/Dialogs/Send.hs
```diff
@@ -1015,8 +1015,37 @@ initiateCrossChainTransfer model networkName envs publicMeta keys fromAccount fr
...
+  :: ( TriggerEvent t m
+     , PerformEvent t m
+     , MonadJSM (Performable m)
+     , HasLogger model t
+     )
+  => model
+  -> [S.ClientEnv]
+  -> Event t (Pact.PactId, Pact.RequestKey)
+  -> m (Event t (Either Text ()))
+pollNodesForSuccess model envs reqEv = pollNodesForReq model envs (first Just <$> reqEv) 12 15 $ \pr mpid->
```
mightybyte: I think we should probably increase the max number of polls a good bit.  Three minutes does seem like a reasonable estimate but I think we should build in more of a buffer.  I'm thinking double it to 24...i.e. 6 minutes.

mightybyte (changes requested):

jmininger:
Closing in favor of #651 due to certain people's branching preferences

PR #650, opened 2021-05-15
Fix Account Details page

emmanueldenloye:
display keyset ref information correctly

### DIFF
--- a/frontend/src/Frontend/UI/Dialogs/AccountDetails.hs
+++ b/frontend/src/Frontend/UI/Dialogs/AccountDetails.hs
@@ -98,7 +98,10 @@ uiAccountDetailsOnChainImpl netname (name, chain, details, account) onClose = Wo
       notesEdit <- notesEditor $ _vanityAccount_notes $ _account_storage account
       pure notesEdit
 
-    let guardTitle = maybe "Keyset" (const "Guard") $ account ^? account_status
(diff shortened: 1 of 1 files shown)


frontend/src/Frontend/UI/Dialogs/AccountDetails.hs
```diff
@@ -117,8 +120,10 @@ uiAccountDetailsOnChainImpl netname (name, chain, details, account) onClose = Wo
               for_ ksKeys $ \key -> uiInputElement $ def
                 & initialAttributes %~ Map.insert "disabled" "disabled" . addToClassAttr "labeled-input__input labeled-input__multiple"
                 & inputElementConfig_initialValue .~ keyToText key
-          AccountGuard_Other g ->
-            void $ displayText (pactGuardTypeText $ Pact.guardTypeOf g) (renderCompactText g) ""
+          AccountGuard_Other g -> case g of
+            (Pact.GKeySetRef (Pact.KeySetName name)) -> do
```
mightybyte, line 124: I wonder if we should abstract over keyset refs at this level and do the work of querying the underlying keyset so we can be more informative to the user.  What do you think @sirlensalot?

mightybyte (approved):

mightybyte (approved):

PR #653, opened 2021-05-27
Get frontend building on ghcjs

jmininger:
- Reliant on unmerged branch of pact that doesn't have `swagger` as a dependency, and exposes `Pact.Types.SigData`
- Fixes kadena-signing-api to no longer be reliant on swagger
- Removes monad-logger and fast-logger and replaces them with our own ghcjs-friendly module in `common`


common/common.cabal
```diff
@@ -13,6 +13,7 @@ library
                , containers
                , dependent-sum
                , dependent-sum-template
+               , fast-logger
```
mightybyte: I'm a little puzzled about how this works because of:

https://github.com/kazu-yamamoto/logger/blob/master/fast-logger/fast-logger.cabal#L38

and

https://github.com/kazu-yamamoto/unix-time/blob/master/unix-time.cabal#L12

Any idea how much work it would take to copy the code and paste it into this repo?

jmininger: Done

mightybyte (approved):
Lovely

PR #655, opened 2021-06-03
V2 migration

emmanueldenloye:


jmininger:
You should make sure that the tests in `frontend/tests/Frontend/VersionedStore*` are changed over from V1 to V2 (or that V2 is added?)

frontend/src/Frontend/VersionedStore/V2.hs
```diff
@@ -0,0 +1,216 @@
...
+  where
+    oldKeysList = maybe [] (IntMap.toList . runIdentity) (DMap.lookup V0.StoreWallet_Keys v0)
+
+    -- We have to walk through the slightly different encoding of the Network information.
+    -- Also if the storage contains _no_ network configuration then we shouldn't break the new version
+    -- by storing an empty object.
+    newNetworks = (\nets -> StoreFrontend_Network_Networks :=> Identity (convertNodeRefs $ V0.unNetworkMap $ runIdentity nets))
+      <$> DMap.lookup V0.StoreNetwork_Networks v0
+
+    -- This will regenerate the missing key. Desktop will recover the key with
```
jmininger, line 90: This is something we really need to look out for with the upcoming chainweaver changes. Might be ok in the immediate future, but we can't copy the same code over for a V2 --> V3 if this ever happens in the future, because then we would risk wiping out users' keys

frontend/src/Frontend/VersionedStore/V2.hs
```diff
@@ -0,0 +1,216 @@
...
+  StoreFrontend_OAuth_Tokens :: StoreFrontend key (Map OAuthProvider AccessToken)
+  StoreFrontend_OAuth_State :: OAuthProvider -> StoreFrontend key OAuthState
+
+  StoreFrontend_Gist_GistRequested :: StoreFrontend key (GistMeta, Text)
+
+  StoreFrontend_ModuleExplorer_SessionFile :: StoreFrontend key Text
+
+deriving instance Show (StoreFrontend key a)
+
+upgradeFromV0 :: (Monad m, HasCrypto key m) => DMap (V0.StoreFrontend key) Identity -> m (DMap (StoreFrontend key) Identity)
```
jmininger, line 61: Might it make more sense to just compose `V1.upgradeFromV0` and `V2.upgradeFromV1` so that we aren't stuck copying code over for each migration and having to make sure that we have all of the "previous stuff" from the intermediary migration as well as all of the new migration stuff?

frontend/src/Frontend/VersionedStore.hs
```diff
@@ -161,3 +168,4 @@ versionedStorage = VersionedStorage
           liftIO $ Api._transactionLogger_rotateLogFile txLogger
           pure ()
         Just (StoreFrontendVersion_1 :=> _) -> pure ()
+        Just (StoreFrontendVersion_2 :=> _) -> pure ()
```
jmininger, line 179: Reminder to fix this

frontend/src/Frontend/Storage/Class.hs
```diff
@@ -169,6 +169,7 @@ restoreLocalStorage
   :: forall storeKeys m
   . ( HasStorage m
     , Monad m
+    , MonadIO m
```
mightybyte: What was the reason for these new `MonadIO` constraints?  Is it left over from debugging?

frontend/src/Frontend/VersionedStore/V2.hs
mightybyte: Hmmmm, interesting point.  I think we probably should compose the migrations otherwise the logical conclusion just won't be scalable.  That's typically the way I've done it with DB migrations.

frontend/tests/Frontend/VersionedStoreSpec.hs
```diff
@@ -292,7 +374,169 @@ test_v0ToV1Upgrade = testCaseSteps "V0 to V1 Upgrade" $ \step -> do
   where
     path = "tests" </> "Frontend" </> "VersionedStoreSpec.files" </> "V0"
 
+fail_test_v0ToV2Upgrade :: FailStorageState -> TestTree
```
mightybyte, line 379: If we use compositional migrations as mentioned above, we might be able to get this test to become more like a property, i.e.

```
migrate a (a+2) == (migrate a (a+1) >> migrate (a+1) (a+2))
```

Not sure if that will be too invasive of a change though.

frontend/src/Frontend/VersionedStore/V2.hs
```diff
@@ -0,0 +1,218 @@
...
+  where
+    dlrep v n
+      | n < 0 = error "fromMultiSet: IMPOSSIBLE"
+      | n == 0 = id
+      | otherwise = (v:) . dlrep v (n - 1)
+
+convertNodeRefs :: Map NetworkName [NodeRef] -> Map NetworkName [NodeRef]
+convertNodeRefs = Map.mapWithKey migrate
+  where
+    migrate = \case
```
mightybyte: Hmmmm...I'm contemplating the possibility of doing the replacement no matter what the network name is.  I think I've seen instances where it's lowercase "mainnet" rather than "Mainnet".  We would have to keep track of which of the node refs were actually found to know which new ref to add.  But that would work for these other cases.

frontend/src/Frontend/Storage/Class.hs
emmanueldenloye: Yeah,  I'm pretty sure that's the reason. I'll take them out.

frontend/tests/Frontend/VersionedStoreSpec.hs
emmanueldenloye: I actually really like this idea. Also adding new tests for new versions will be a lot easier to make.

frontend/src/Frontend/VersionedStore/V2.hs
```diff
@@ -0,0 +1,216 @@
...
+
+  StoreFrontend_Gist_GistRequested :: StoreFrontend key (GistMeta, Text)
+
+  StoreFrontend_ModuleExplorer_SessionFile :: StoreFrontend key Text
+
+deriving instance Show (StoreFrontend key a)
+
+upgradeFromV0 :: (Monad m, HasCrypto key m) => DMap (V0.StoreFrontend key) Identity -> m (DMap (StoreFrontend key) Identity)
+upgradeFromV0 v0 = do
+  (newKeysList, newAccountStorage) <- foldMapM splitOldKey oldKeysList
```
jmininger, line 63: `splitOldKey` really concerns me, for the web version. Its relying on the keys' index in a list, and then using that index to "regenerate" them instead of saving /copying them somewhere. This code is so easy to accidentally copy over, for, say the next migration. We should either rewrite it or at the very least use CPP to make it a no-op on ghcjs. Not quite sure its in the scope of this task, but it could so easily go really bad really quickly if it isn't taken care of imo

frontend/src/Frontend/VersionedStore/V2.hs
```diff
@@ -152,16 +155,19 @@ fromMultiSet = ($ []) . Map.foldrWithKey (\k i -> (.) (dlrep k i)) id
       | otherwise = (v:) . dlrep v (n - 1)
 
 convertNodeRefs :: Map NetworkName [NodeRef] -> Map NetworkName [NodeRef]
-convertNodeRefs = Map.mapWithKey migrate
+convertNodeRefs = fmap migrate
   where
```
jmininger, line 177: Is the nested `where` necessary ?

frontend/src/Frontend/VersionedStore/V2.hs
```diff
@@ -153,32 +153,37 @@ fromMultiSet = ($ []) . Map.foldrWithKey (\k i -> (.) (dlrep k i)) id
...
-  That f <*> These b x = These b (f x)
-  These a f <*> That x = These a (f x)
-  These a f <*> These b x = These (a <> b) (f x)
+-- It is equivalent to "These () a" except that we take out the
+-- "This" constructor. With it gone, we still get the same applicative
+-- operations from "These" (sans "This") but we don't have to account for
+-- the "This" constructor in situations where we know it is impossible to
+-- produce it. This is also morally equivalent to Either () b with a
+-- different applicative instance
+data Deez b = Dis b | Dat b
```
emilypi: slightly misleading comment. 

```haskell
These a b 
~ a + b + a*b
```

But

```haskell
These () a 
~ 1 + a + 1*a 
~ 1 + a + a 
~ Maybe (Either a a)
```

and 

```haskell
Either () b 
~ 1 + b
~ Maybe b
```

You want a newtype for `Either a a` here to get the modified `Applicative`.

PR #656, opened 2021-06-03
Fix server deploy functions

jmininger:
Some of our nix deployment infrastructure has gone stale after bumping obelisk. Our `server` func is mostly copied over from the obelisk one, but we overload the configuration with an extra `activationScript` (and at one point had our own pact server module that we added as well). This customization means we can't just use the default one supplied by ob


mightybyte (approved):
Ahh, I see.  Nice sleuthing!

Issue #665, opened 2021-07-29
Leading decimals aren't accepted using transfer widget in amount field


jmininger:
An amount of `.5001` will not be accepted (but does not issue a warning to the user about why), but an amount of `0.5001` will be

jmininger:
2.2.0 has been released. This issue should be fixed, so im closing this. Feel free to reopen if the issue persists

PR #666, opened 2021-08-02
Fix decimal-related bugs

mightybyte:


frontend/src/Frontend/UI/Form/Common.hs
```diff
@@ -92,14 +92,15 @@ amountFormWidget
...
-  where
-    p t = case readMaybe (T.unpack t) of
-            Nothing -> Left "Not a valid number"
-            Just x
-              | x < 0 -> Left "Cannot be negative"
-              | D.decimalPlaces x > maxCoinPrecision -> Left "Too many decimal places"
-              | otherwise -> Right x
+    parsingFormWidget parseAmount (either (const "") tshow) cfg
+
+parseAmount :: Text -> Either String Decimal
```
jmininger, line 97: Can we add a check for a leading decimal here, and then concat the value with `0` if this is the case? This would allow us to handle amounts like `.4` which we currently don't handle (we also don't give a failure reason, which is really annoying to have to figure out for the avg user)

jmininger: Added it myself

jmininger:
This PR also addresses #665

Issue #670, opened 2021-08-13
Trying to initiate cross chain transfer to an account with module as guard keeps trying to load keyset


floordebest:
When trying to make a cross chain transfer to an account that doesn't have a keyset as guard and balance of 0.0 (in my case a gaspayer account that has a gasstation module as guard) the pop up shows "Querying keysets..." and all buttons are unresponsive, clicking outside the popup wont close it.

sender, chain 0 (testnet): fw-gaspayer (keyset as guard)
receiver, chain 1 gaspayer account (testnet): fw-gaspayer-v2
Gasstation module: free.floppie-wallet-gas-station

floordebest:
Running on Macbook Pro / MacOs Big Sur 11.5 / Chainweaver 2.1.1

jmininger:
I think this an incomplete pattern match on guard type that snuck in. I ran across it the other day. Should be fixed in the next release. Thank you for raising it.

jmininger:
2.2.0 has been released. This issue should be fixed, so im closing this. Feel free to reopen if the issue persists

Issue #652, opened 2021-05-23
Add support for gas station


Thanos420NoScope:
Chainweaver could look up the receiving account balance and enter the public gas station if its empty or doesnt exist
![image](https://user-images.githubusercontent.com/6309193/119244918-1be30d80-bb43-11eb-99a3-648c6fde7af2.png)
Could also be entered by default on crosschains to simplify

jmininger:
2.2.0 has been released. This issue should be fixed, so im closing this. Feel free to reopen if the issue persists

PR #673, opened 2021-08-16
free-x-chain gas station default

jmininger:
Adds `free-x-chain-gas` as the default destination-chain gas payer for accounts that don't yet exist (should i also add for accounts with 0 balance?) . Also sets the gasLimit for these txs at 400. 

Addresses #652 

Additionally patches an incomplete pattern match that was causing a runtime exception, although doesn't exactly _solve_ the underlying problem of selecting a guard parameter for cross-chain transfers to vanity accounts (that already exist) . The most common scenario here is when doing a cross-chain transfer to a gas station. See #670

### DIFF
--- a/frontend/src/Frontend/UI/Transfer.hs
+++ b/frontend/src/Frontend/UI/Transfer.hs
@@ -1020,13 +1020,15 @@ gasPayersSection
   => model
   -> SharedNetInfo NodeInfo
   -> Map AccountName (AccountStatus AccountDetails)
+  -> Map AccountName (AccountStatus AccountDetails)
   -> TransferInfo
   -> m (GasPayers t)
-gasPayersSection model netInfo fks ti = do
+gasPayersSection model netInfo fks tks ti = do
     let fromChain = _ca_chain (_ti_fromAccount ti)
         fromAccount = _ca_account (_ti_fromAccount ti)
+        toAccount = _ca_account (_ti_toAccount ti)
         toChain = _ca_chain (_ti_toAccount ti)
-    (dgp1, mdgp2) <- if fromChain == toChain
+    (dgp1, mdmgp2) <- if fromChain == toChain
       then do
         let initialGasPayer = if _ti_maxAmount ti then Nothing else Just (_ca_account $ _ti_fromAccount ti)
         let goodGasPayer a = if _ti_maxAmount ti && a == fromAccount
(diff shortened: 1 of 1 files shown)


frontend/src/Frontend/UI/Transfer.hs
```diff
@@ -562,9 +562,11 @@ checkReceivingAccount model netInfo ti ty fks tks fromPair = do
...
         if (_ca_chain $ _ti_fromAccount ti) /= (_ca_chain $ _ti_toAccount ti)
           then do
-            let AccountGuard_KeySetLike (KeySetHeritage ks p _ref) = g
-            let ti2 = ti { _ti_toKeyset = Just $ UserKeyset ks (parseKeysetPred p) }
-            transferDialog model netInfo ti2 ty fks tks fromPair
+            case g of
+              AccountGuard_KeySetLike (KeySetHeritage ks p _ref) ->
+                let ti2 = ti { _ti_toKeyset = Just $ UserKeyset ks (parseKeysetPred p) }
+                in transferDialog model netInfo ti2 ty fks tks fromPair
+              _ -> transferDialog model netInfo ti ty fks tks fromPair
```
mightybyte: Can you add a comment associated with this catch-all case with a little background?  Something about the complexity of determining from an arbitrary guard what keys are needed for signing, and maybe mentioning the super generic big hammer of allowing the user to add arbitrary signing keys...

frontend/src/Frontend/UI/Transfer.hs
```diff
@@ -1037,40 +1041,57 @@ gasPayersSection model netInfo fks ti = do
         pure $ (dgp1, Nothing)
       else do
         let mkLabel c = T.pack $ printf "Gas Paying Account (Chain %s)" (T.unpack $ _chainId c)
+            destinationStatus = Map.lookup toAccount tks
+            defaultDestGasPayer = case destinationStatus of
+              Just (AccountStatus_Exists _) -> toAccount
```
mightybyte: I like your idea of checking for a sufficient balance.  Seems like a good thing to add while we're here.

frontend/src/Frontend/UI/Transfer.hs
```diff
@@ -1037,40 +1041,57 @@ gasPayersSection model netInfo fks ti = do
...
-    case mgp2Debounced of
-      Nothing -> return $ GasPayers gp1 Nothing
-      Just dgp2 -> do
-        gp2Keys <- networkHold (return never) (getGasPayerKeys toChain <$> dgp2)
+    --Outer maybe represents single chain transfer (so no second gas payer). Inner maybe represents
+    --an account name that is less than 3 chars for the "Destination Gas Payer" input field
+    mgp2 <- forM mdmgp2 $ \dmgp2 -> do
+        -- Take the initial value of the destination gas payer and lookup its details
+        initVal <- sample $ current dmgp2
+        initGasPayerDetails <- (fmap . fmap) (GasPayerDetails toChain initVal) $
```
mightybyte: We have `<$$>` and `<$$$>` convenience functions that I think read more nicely than the composed`fmap`.

frontend/src/Frontend/UI/Transfer.hs
```diff
@@ -1037,40 +1041,57 @@ gasPayersSection model netInfo fks ti = do
...
       ]
-    case mgp2Debounced of
-      Nothing -> return $ GasPayers gp1 Nothing
-      Just dgp2 -> do
-        gp2Keys <- networkHold (return never) (getGasPayerKeys toChain <$> dgp2)
+    --Outer maybe represents single chain transfer (so no second gas payer). Inner maybe represents
+    --an account name that is less than 3 chars for the "Destination Gas Payer" input field
+    mgp2 <- forM mdmgp2 $ \dmgp2 -> do
+        -- Take the initial value of the destination gas payer and lookup its details
+        initVal <- sample $ current dmgp2
```
mightybyte: What's the reason for using `sample` here?  I have a general heuristic that `sample` is often a bad idea.  It carries with it the big question mark of when the sample occurs and is usually more strict than you actually want.  Using `networkHold` or `networkView` tend to make it a lot more clear what event is triggering the action.

mightybyte (changes requested):

frontend/src/Frontend/UI/Transfer.hs
jmininger: I added a `balance > 0` check because anything else seems absurdly tricky

frontend/src/Frontend/UI/Transfer.hs
jmininger: Switched to just use the initial default value for the input-field

mightybyte (approved):

PR #680, opened 2021-08-23
Propagate GasLimit to the second part of cross-chain transfer properly

jmininger:
Also commented out a lot of dead code

I see this as more of a temporary patch to get stuff working. I think ultimately we should aim to remove the centralized state (what even is a "Current Chainweaver PublicMeta" structure ? And why is every PublicMeta form able to set it?)

### DIFF
--- a/frontend/src/Frontend/UI/Transfer.hs
+++ b/frontend/src/Frontend/UI/Transfer.hs
@@ -646,7 +646,6 @@ transferDialog model netInfo ti ty fks tks _ = do
       (currentTab, _done) <- transferTabs newTab
       (conf, meta, payload, dSignedCmd, destChainInfo) <- mainSection currentTab
       (cancel, newTab, next) <- footerSection currentTab meta dSignedCmd
-
     case destChainInfo of
       Nothing -> do
         let nextScreen = ffor (tag (current dSignedCmd) next) $ \case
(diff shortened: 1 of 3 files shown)


frontend/src/Frontend/UI/Transfer.hs
```diff
@@ -835,7 +834,8 @@ crossChainTransferAndStatus model netInfo ti cmd mdestGP destSigners = Workflow
             el "p" $ text $ "Cross chain transfer initiated on chain " <> _chainId fromChain
 
           keys <- sample $ current $ model ^. wallet_keys
-          runUnfinishedCrossChainTransfer logL netInfo keys fromChain toChain mdestGP rk
+          toChainMeta <- sample $ flip transferMetaToPublicMeta toChain <$> current meta
```
mightybyte: Instead of passing this in as a Dynamic and then calling sample on it, can we just pass in a pure TransferMeta?

mightybyte (approved):

frontend/src/Frontend/UI/Transfer.hs
jmininger: Well, it's not pure. The GasLimit and other metadata comes from the input form, and can be set to anything the user wants

mightybyte: Oh right, I missed that bit.

frontend/src/Frontend/UI/Transfer.hs
```diff
@@ -660,7 +659,7 @@ transferDialog model netInfo ti ty fks tks _ = do
         let nextScreen = ffor (tag allDyns next) $ \case
               (_,Nothing,_,_) -> Workflow $ pure (mempty, never)
               (p,Just sc,gp,ss) -> previewDialog model netInfo ti payload sc $
-                                     crossChainTransferAndStatus model netInfo ti sc gp ss
+                                     crossChainTransferAndStatus model netInfo ti sc gp ss meta
```
mightybyte: Instead of doing `sample` inside the `crossChainTransferAndStatus` function I think we should add those dyns to `allDyns` a few lines up because that's the place where we're getting our event that determines when all these values should be sampled.

jmininger: Good catch. Done

mightybyte (approved):

PR #683, opened 2021-08-25
Unpad ChainIds when passing them around

jmininger:
We were using padded chain ids so that they would sort properly, but we never unpadded them before propagating them back through our app. This resulted in using the wrong lens to persist VanityAccountNotes which resulted in those notes not persisting after pressing the "Done" button

### DIFF
--- a/frontend/src/Frontend/UI/Wallet.hs
+++ b/frontend/src/Frontend/UI/Wallet.hs
@@ -345,7 +345,7 @@ uiAccountItem cwKeys startsOpen name accountInfo = do
       False -> pure never
       True -> trAcc $ do
         td blank -- Arrow column
-        td $ text $ "Chain " <> _chainId paddedChain
+        td $ text $ "Chain " <> _chainId chain
         td $ dynText $ maybe "" ownershipText <$> getAccountOwnership cwKeys details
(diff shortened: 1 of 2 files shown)


frontend/src/Frontend/UI/Wallet.hs
```diff
@@ -334,7 +345,7 @@ uiAccountItem cwKeys startsOpen name accountInfo = do
       False -> pure never
       True -> trAcc $ do
         td blank -- Arrow column
-        td $ text $ "Chain " <> _chainId chain
+        td $ text $ "Chain " <> _chainId paddedChain
```
mightybyte: Do we want `paddedChain` here?  Seems like we would want to show the user the unpadded value?

jmininger: For displaying, it actually kinda looks a bit nicer because they are all the same width. Although you could also make the arg that its ultimately misinformation that could lead to confusing a user somehow?

mightybyte: Yeah, I think I don't want to mislead the user.

jmininger: Done

mightybyte (approved):

PR #688, opened 2021-10-02
Add a working ca-certificate.crt for linux apps to use

jmininger:
The default one contained an expired certificate and would break all https requests to api.chainweb.com


larskuhtz (approved):

larskuhtz:
If we merge this we may also create an issue to remind us to eventually do the proper fix, which is, updating the gnutls package version.

mightybyte (approved):

PR #695, opened 2021-10-25
Add warning for transfers to accounts with names that don't match 🔑 set

jmininger:


common/src/Common/Wallet.hs
```diff
@@ -396,6 +403,16 @@ filterKeyPairs :: Set PublicKey -> IntMap (Key key) -> [KeyPair key]
...
   where toMap = Map.fromList . fmap (\k -> (_keyPair_publicKey $ _key_pair k, _key_pair k)) . IntMap.elems
 
+-- Checks that accName = pubkey has a keys-all guard with the same pubkey
+accountNameMatchesKeyset :: AccountName -> AccountGuard -> Bool
+accountNameMatchesKeyset accName g = case g of
+  AccountGuard_Other _ -> False
+  AccountGuard_KeySetLike (KeySetHeritage ksKeys ksPred _ksRef) ->
+    case parsePubKeyOrKAccount accName of
+      -- Should we allow k:<pubkey> to be True?
+      (_, Right pk) -> ksPred == defaultPredicate && ksKeys == Set.singleton pk
```
mightybyte: I don't think we should depend on predicate here.  The logic seems strongest when applied irrespective of predicate.

jmininger: addressed

mightybyte (approved):

PR #698, opened 2021-11-03
Fungible support

jmininger:


mightybyte:
I think the gas check for the sender on the sending chain might be wrong.

jmininger:
Close in favor of #767

PR #710, opened 2021-12-16
add k-account button on keys page

jmininger:


frontend/src/Frontend/UI/Button.hs
```diff
@@ -347,6 +348,11 @@ transferToButton cfg =
     imgWithAltCls "button__text-icon" (static @"img/transfer-to.svg") "Transfer to" blank
     elClass "span" "button__text button__text-exclusive" $ text "Transfer to"
 
+addKAccountButton :: StaticButtonConstraints t m => UiButtonCfg -> m (Event t ())
+addKAccountButton cfg =
+  uiButton (cfg & uiButtonCfg_class <>~ "button_type_secondary" <> "button_type_secondary") $
+    text "Add K Account"
```
mightybyte: I wonder if we should say `Add k: Account` instead of `Add K Account`.  Not sure though...I'm on the fence...

mightybyte (approved):

mightybyte:
Looks great!

frontend/src/Frontend/UI/Button.hs
jmininger: Maybe "Add Single Signer Account"

PR #711, opened 2021-12-17
Import and Export wallet functionality for web app

dfordivam:


frontend/src/Frontend/Setup/ImportExport.hs
```diff
@@ -100,36 +108,50 @@ extractImportVersionField
 extractImportVersionField =
   extractImportDataField
 
+data ImportWidgetApis bipStorage key n m = ImportWidgetApis
+  { _importWidgetApis_bipStorageKey :: bipStorage key
+  , _importWidgetApis_pwCheck :: (key -> Password -> m Bool)
+  , _importWidgetApis_runF :: (forall a . () => key -> Password -> n m a -> m a)
```
dfordivam, line 113: This could be given a better name, like `runCryptoT`

PR #712, opened 2021-12-17
Generate QR code for signatures

mightybyte:


jmininger:
Looks good. Any reason its still in draft?

Issue #636, opened 2020-08-26
Ubuntu 20.04 [ getLoginName: does not exist (No such device or address) ] when program is ran


chinatimmy:
Using Xubuntu 20.04.1 LTS x86_64 

**ran command** 
redacted@redacted:/usr/libexec/chainweaver$ ./kadena-chainweaver 

**output produced**
Executable path: /usr/libexec/chainweaver/
kadena-chainweaver: getLoginName: does not exist (No such device or address)

**recommendation/enhancement**
App image or Flatpak as some development teams don't all use the same distro

jmininger:
Spent a bit of time looking into this...The following thread seems relevant, along with the linked SO q/a: 
https://gitlab.haskell.org/ghc/ghc/-/issues/1487
https://stackoverflow.com/questions/4785126/getlogin-c-function-returns-null-and-error-no-such-file-or-directory

with the chainweaver code here, being the chunk in question: https://github.com/kadena-io/chainweaver/blob/develop/linux/src/Linux.hs#L178

freQniK:
Yeah, I'm getting the same thing on recent build still. Ubuntu 20.04. Not a very good wallet if it doesn't even load.

trendzetter:
https://discord.com/channels/502858632178958377/631965054778212362/948882072490098708

vladimir-litvinenko:
@trendzetter @freQniK  Are there any progress on merging this? Just installed .deb on Ubuntu. I'm experiencing the same issue:

```
 > sudo dpkg -i kadena-chainweaver-linux-2.2.2.0.deb 
(Reading database ... 704403 files and directories currently installed.)
Preparing to unpack kadena-chainweaver-linux-2.2.2.0.deb ...
Unpacking kadena-chainweaver (2.2.2.0) over (2.2.2.0) ...
Setting up kadena-chainweaver (2.2.2.0) ...
Processing triggers for gnome-menus (3.36.0-1ubuntu1) ...
Processing triggers for desktop-file-utils (0.24+linuxmint1) ...
Processing triggers for mime-support (3.64ubuntu1) ...

 > kadena-chainweaver 
Executable path: /usr/libexec/chainweaver/
kadena-chainweaver: getLoginName: does not exist (No such device or address)
 
```

PR #717, opened 2021-12-19
Fix Issue #636 on Ubuntu

freQniK:
instead of getLoginName, use getEffectiveUserName to properly get the name of the user associated to the logged in User ID

see:  https://github.com/kadena-io/chainweaver/issues/636


jmininger:
Thanks for giving this a go. Do you think you could make it so that it branches on the failure of the original piece of code and only runs the newer version if the former fails? Thanks!

freQniK:
Yeah, I'll see about adding the proper exception handling. Give me a few days.

PR #720, opened 2021-12-27
Added support for signing SigData JSON in Keys section

ohri-anurag:
This PR contains the code for:
1. Adding an input text box for receiving SigData JSON from the user
2. If everything's correct, show the signed hash of the input in another text box, along with a copy button
3. If there are errors during parsing or if the hash was incorrect, show them below the input box.


ohri-anurag:
This PR is superseded by another PR: https://github.com/kadena-io/chainweaver/pull/725

PR #725, opened 2022-01-03
Add JSON/YAML signing to Keys section

ohri-anurag:
I have added support for 3 scenarios here, as per my discussion with @jmininger on Slack. This PR effectively supersedes my older PR, which I will mark as abandoned.

The three scenarios are:
1. User enters `JSON` -> We parse `SigData Text` -> We produce the signature.
2. User enters `JSON` -> We parse `Payload PublicMeta Text` -> We produce a `Command Text`.
3. User enters `YAML` -> We parse `SigData Text` -> We produce another `SigData Text`, but this one contains the signature of the current user.

I have also added support for showing errors. A copy button is also present whenever there is no error, and the output is non-empty.


mightybyte (commented):
Looks great!  A few small comments...

1. Now that we have this safer signing mechanism, I think we should probably remove the existing signing since it is very unsafe.
2. Since we now have a ghcjs-compatible yaml library, let's switch this code https://github.com/kadena-io/chainweaver/blob/a4ae0aee0c6db5ab9b36691a3adef47a60b59ecf/frontend/src/Frontend/UI/Transfer.hs#L1467 to use it and eliminate the use of the `yaml` library.

PR #727, opened 2022-01-09
Sig Builder

jmininger:


frontend/src/Frontend/Network.hs
```diff
@@ -69,6 +69,9 @@ module Frontend.Network
   , getNetworkNameAndMeta
   , getCreationTime
   , encodeAsText
+
+-- TODO
```
mightybyte: This TODO can probably be removed.

frontend/src/Frontend/UI/Dialogs/SigBuilder.hs
```diff
@@ -0,0 +1,735 @@
...
+     )
+  => SigData Text
+  -> [KeyPair key]
+  -- ^ Keys which we are signing with
+  -> [(PublicKeyHex, UserSig)]
+  -> m (SigData Text)
+addSigsToSigData sd signingKeys sigList = do
+  --TODO: Double check hash here and fail if they dontmatch
+  let -- cmd = encodeAsText $ encode payload
+      -- cmdHashL = hash (T.encodeUtf8 cmd)
```
mightybyte: Was this meant to be uncommented?

frontend/src/Frontend/UI/Dialogs/SigBuilder.hs
```diff
@@ -0,0 +1,735 @@
...
+  -> [KeyPair key]  -- ChainweaverKeys
+  -> Workflow t m (Event t ()) -- Workflow for going back
+  -> [(PublicKeyHex, UserSig)] -- User-supplied Sigs
+  -> Workflow t m (Event t ())
+signAndShowSigDialog model sbr keys backW sigsOrPrivate = Workflow $ mdo
+  onClose <- modalHeader $ text "Sig Data"
+  -- This allows a "loading" page to render before we attempt to do the really computationally
+  -- expensive sigs
+  pb <- delay 0.1 =<< getPostBuild
+  -- TODO: Can we forkIO the sig process and display something after they are done?
```
mightybyte, line 543: Maybe, but that seems like it would present an avenue of attack.  I would probably stick with the conservative "no signatures are computed until the user has give explicit approval" philosophy.

frontend/src/Frontend/UI/Dialogs/SigBuilder.hs
```diff
@@ -0,0 +1,735 @@
...
+txnInputDialog model mInitVal = Workflow $ mdo
+  onClose <- modalHeader $ text "Signature Builder"
+  let cwKeys = IMap.elems <$> (model^.wallet_keys)
+      selNodes = model ^. network_selectedNodes
+      nid = (fmap (mkNetworkName . nodeVersion) . headMay . rights) <$> selNodes
+      keysAndNet = current $ (,) <$> cwKeys <*> nid
+  dmSigData <- modalMain $ divClass "group" $ parseInputToSigDataWidget mInitVal
+  (onCancel, approve) <- modalFooter $ (,)
+    <$> cancelButton def "Cancel"
+    <*> confirmButton (def & uiButtonCfg_disabled .~ ( isNothing <$> dmSigData )) "Approve"
```
mightybyte: Let's change this to say "Review"

mightybyte (approved):

frontend/src/Frontend/UI/Dialogs/SigBuilder.hs
jmininger: So the idea was still to fork after the user presses "Accept / Sign". But i wanted finer grain changes so we could present a better loading screen like: " Loading: 3/8 signatures added" -- instead of just being stuck on it as it adds all 8 sigs at once

PR #734, opened 2022-02-01
Persistent fungibles

jmininger:


jmininger:
Close in favor of #767

PR #736, opened 2022-02-10
Removed ghcjs specific code

ohri-anurag:
Removed the dependency on juicypixels as a whole, by keeping the `encodeBitmap` function from juicypixels into our fork of `qrcode-juicypixels`(maybe we should rename this, since it is not dependent on juicypixels anymore).


ohri-anurag:
@jmininger I forgot to tag you here.

PR #740, opened 2022-02-23
GHCJS dedupe flag works

ohri-anurag:
The issue with adding the dedupe flag, along with Google closure optimization was the clashing of function names. Chainweaver has a lot of static js files, that define objects in the global namespace.

Some of these names are also generated by the closure compiler with the dedupe flag on, namely `ace` and `lib`. Due to these naming issues, the JS static libraries generated errors.

We have added an `externs.js` file, which we pass on to Google Closure here: https://github.com/obsidiansystems/obelisk/commit/93fb8d1ec612b63ce9d6c38eeecf314fde8ae9fc.
This is why this branch uses a different branch of obelisk.

### DIFF
--- a/.obelisk/impl/github.json
+++ b/.obelisk/impl/github.json
@@ -3,6 +3,6 @@
   "repo": "obelisk",
   "branch": "ao-kadena-externs",
   "private": false,
-  "rev": "93fb8d1ec612b63ce9d6c38eeecf314fde8ae9fc",
-  "sha256": "1h55amakh75v9zsfb1zcj2g50h1ghvy0qnm3105bw30zwjpn0mfd"
+  "rev": "b2264f41b1d600242882c76d3cbfaaf0820f3ca5",
+  "sha256": "0hwxla2p36s6ipy6vfw4y3wxf70wfrgbkm915slgdcdyy1ph5952"
 }
--- a/.obelisk/impl/github.json
+++ b/.obelisk/impl/github.json
@@ -1,7 +1,8 @@
 {
   "owner": "obsidiansystems",
   "repo": "obelisk",
-  "branch": "master",
-  "rev": "7ad33cbe3e84b209e83c505ce25486445bbd602e",
-  "sha256": "0dlk8y6rxc87crw7764zq2py7nqn38lw496ca1y893m9gdq8qdkz"
+  "branch": "ao-kadena-externs",
+  "private": false,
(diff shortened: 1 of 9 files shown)


obApp.nix
```diff
@@ -19,7 +19,7 @@ in with obelisk;
     # ios.bundleIdentifier = "systems.obsidian.obelisk.examples.minimal";
     # ios.bundleName = "Obelisk Minimal Example";
 
-    __closureCompilerOptimizationLevel = "SIMPLE";
+    # __closureCompilerOptimizationLevel = "SIMPLE";
```
ohri-anurag: @jmininger This line is unnecessary, since simple is the default. I included it as a comment so that it matched with Doug's commit.

frontend/frontend.cabal
```diff
@@ -257,8 +257,8 @@ executable frontend
   -- https://github.com/ghcjs/ghcjs/issues/748 is fixed. -dedupe greatly
```
dfordivam: Update the comment?

frontend/src/Frontend.hs
```diff
@@ -139,6 +139,7 @@ newHead routeText = do
   ss "https://fonts.googleapis.com/css?family=Work+Sans"
   ss (static @"css/font-awesome.min.css")
   ss (static @"css/ace-theme-chainweaver.css")
+  js "/static/externs.js"
```
dfordivam: Is it really necessary to import this file in the head? I thought it is is only needed as an input to the closureCompiler step

obApp.nix
dfordivam: I think this could be removed

frontend/src/Frontend.hs
ohri-anurag: Removed.

.obelisk/impl/github.json
```diff
@@ -1,7 +1,8 @@
 {
   "owner": "obsidiansystems",
   "repo": "obelisk",
-  "branch": "master",
-  "rev": "7ad33cbe3e84b209e83c505ce25486445bbd602e",
-  "sha256": "0dlk8y6rxc87crw7764zq2py7nqn38lw496ca1y893m9gdq8qdkz"
+  "branch": "ao-kadena-externs",
```
jmininger, line 4: Need to fork and point to kadena-io repo

jmininger:
Closing in favor of #769

PR #742, opened 2022-02-23
Reduce concurrency without obelisk bump

meditans:
These are the same changes done for reducing the amount of XHR to solve the freeze problem on newer versions of macOS but without the obelisk bump. However, there's still a minor upgrade in kpkgs, needed to fix the build (relaxing some tests in libuv).

### DIFF
--- a/dep/kpkgs/github.json
+++ b/dep/kpkgs/github.json
@@ -1,8 +1,8 @@
 {
-  "owner": "kadena-io",
+  "owner": "obsidiansystems",
   "repo": "kpkgs",
-  "branch": "master",
+  "branch": "cn/remove-more-tests-from-libuv",
   "private": false,
-  "rev": "763f1763207a02fbf6257bd8e2a8cef2a9f86e1d",
-  "sha256": "1dncp11hn65dvwpqrw3di6hdlykjpgc0w0ii5lvwdziac861ak2n"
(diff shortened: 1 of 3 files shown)


.obelisk/impl/github.json
```diff
@@ -1,7 +1,8 @@
 {
   "owner": "obsidiansystems",
   "repo": "obelisk",
-  "branch": "master",
-  "rev": "7ad33cbe3e84b209e83c505ce25486445bbd602e",
-  "sha256": "0dlk8y6rxc87crw7764zq2py7nqn38lw496ca1y893m9gdq8qdkz"
+  "branch": "cn/ignore-more-libuv-tests-obelisk-side",
```
jmininger: Can we just add this change: https://github.com/obsidiansystems/nixpkgs/commit/c2304d2565502216d65fd68660c39784fe253104 to an overlay somwhere more local (obApp.nix or kpkgs perhaps ?)

jmininger: Can we just add this change: https://github.com/obsidiansystems/nixpkgs/commit/c2304d2565502216d65fd68660c39784fe253104 to an overlay somwhere more local (obApp.nix or kpkgs perhaps ?)

jmininger:
Closing in favor of #754

PR #745, opened 2022-03-01
Various SigBuilder

jmininger:
- Adds a preview that shows result + gas cost when all signatures exist in SigBuilder
- Adds QRcodes for hash + sigdata in SigBuilder 
- Changes "Add Account" button to "Watch Account" 
- Changes Deploy workflow's "Accounts" section to "Signers" which was mislabeled

### DIFF
--- a/frontend/src/Frontend/UI/Dialogs/AddVanityAccount.hs
+++ b/frontend/src/Frontend/UI/Dialogs/AddVanityAccount.hs
@@ -31,7 +31,7 @@ uiWatchAccountButton
   -> m mConf
 uiWatchAccountButton m = do
   eOpenAddAccount <- uiButton headerBtnCfg  (text "+ Watch Account")
-  pure $ mempty & modalCfg_setModal .~ (Just (uiWatchAccountDialog m) <$ eOpenAddAccount)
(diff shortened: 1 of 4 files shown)


mightybyte (approved):

frontend/src/Frontend/UI/Dialogs/AddVanityAccount.hs
```diff
@@ -22,15 +22,15 @@ import Frontend.UI.Widgets.Helpers (dialogSectionHeading)
...
   :: forall t m key mConf
      . ( MonadWidget t m, Monoid mConf
        , HasModalCfg mConf (ModalImpl m key t) t
        )
   => ModalIde m key t
   -> m mConf
-uiAddAccountButton m = do
-  eOpenAddAccount <- uiButton headerBtnCfg  (text "+ Add Account")
+uiWatchAccountButton m = do
+  eOpenAddAccount <- uiButton headerBtnCfg  (text "+ Watch Account")
```
enobayram, line 33: Seems like [this line](https://github.com/kadena-io/chainweaver/blob/446544ba607922e6dbed38bb8f8c301c1f49a0f7/frontend/src/Frontend/UI/Wallet.hs#L183) needs to change as well. Without that, the No Accounts Found message refers to an "+ Add Account button".

image

PR #746, opened 2022-03-03
Added an input widget that takes monadic validation functions

ohri-anurag:
I have made two changes in this PR:
1. Added a new widget, its name is similar to the existing one, with `async` appended at the end. It differs from the existing function:
    - It takes a validator function that returns a new type, called `ValidationResult`.
      - Failure
      - Success
      - Warning
    - The Warning constructor is treated as an acceptable value, along with a warning to show.
    - Note that this type allows us to show error messages for valid accounts, which was not possible with the either type.
2. I also made changes to `uiInputWithPopover`:
    - Added a new state for loading, in this state the loading gif is shown.
    - For this function, I did not create a new variant.
    - This widget receives a function that returns an `Event` of states.
    - Already existing widgets will work as they were, since they are not making use of the new state. The `Event` that they pass to `uiInputWithPopover` will not have the loading state.


frontend/src/Frontend/UI/Widgets.hs
```diff
@@ -1108,6 +1110,36 @@ uiAccountNameInput label inlineLabel initval onSetName validateName = do
...
+     , PerformEvent t m
+     , TriggerEvent t m
+     , MonadJSM (Performable m)
+     , MonadFix m
+     )
+  => Text                     -- ^ The label for this input
+  -> Bool                     -- ^ Should the label be inline?
+  -> Maybe Text               -- ^ Initial Value for the input
+  -> Event t (Maybe Text)     -- ^ An event that can be used to set the input's value
+  -> Dynamic t (Event t Text -> m (Event t (ValidationResult Text a)))
```
jmininger: What is the point of this being Dynamic?

ohri-anurag: Removed.

jmininger (approved):

PR #749, opened 2022-03-11
Gas payer validation

ohri-anurag:
**NOTE**: This PR should follow this one: #746. Please review and merge that PR before reviewing this one.

This PR contains changes for:
1. Validating the gas payer. User must have input a valid gas payer for them to proceed in the Transfer Dialog.
2. Showing a warning error, and a warning tab, when:
  a. The transaction is a max-amount transaction
  b. The sending account is also the gas payer.
3. In case the conditions described in point 2 are met, the amount that will effectively be transferred will be (max-amount - gas fees). Show this in the warning tab, and make it so that the effective amount is transferred.


frontend/src/Frontend/UI/TabBar.hs
```diff
@@ -62,6 +64,17 @@ data TabBarCfg t m tab = TabBarCfg
 
 makePactLenses ''TabBarCfg
 
+-- | Tab bar configuration, with dynamic tabs
+data TabBarDynCfg t m tab = TabBarDynCfg
```
jmininger: Instead of duplicating this stuff, can we just use `TabBarDynCfg` for everything and use `constDyn` to retrofit the current users of TabBarCfg with the new dyn config?

frontend/src/Frontend/UI/Transfer.hs
```diff
@@ -678,61 +678,83 @@ transferDialog model netInfo ti ty fks tks unused = Workflow $ do
...
+      _ti_maxAmount currentTi && Just (_ca_account $ _ti_fromAccount currentTi) == _transferMeta_senderAccount currentMeta
     mainSection currentTab = elClass "div" "modal__main" $ do
-      (conf, meta, destChainInfo) <- tabPane mempty currentTab TransferTab_Metadata $
+      (conf, meta, destChainInfo, tiDyn) <- tabPane mempty currentTab TransferTab_Metadata $
         transferMetadata model netInfo fks tks ti ty
-      let payload = buildUnsignedCmd netInfo ti ty <$> meta
+      void $ tabPane mempty currentTab TransferTab_MaxTransactionWarning $ do
+        el "h2" $ text "You have asked for a max transfer but used the same account to pay gas. As such, the amount being transferred has changed."
+        el "div" $ do
+          el "span" $ text "Amount being transferred (after deducting gas): "
```
jmininger: Can we add an: "Original Amount" and "Gas Cost" in addition to "New Amount" ?

ohri-anurag: Done.

frontend/src/Frontend/UI/TabBar.hs
ohri-anurag: Done.

jmininger (approved):

jmininger (changes requested):
Gas payer validation functionality broken. Do not merge

jmininger:
Closing in favor of #766

PR #755, opened 2022-03-23
Quicksign v2

jmininger:
Supports the new `CommandSigData` type, and displays previously existing signatures collected in both `SigBuilder` and new `Quicksign` interfaces


dep/signing-api/github.json
```diff
@@ -1,8 +1,8 @@
 {
   "owner": "kadena-io",
   "repo": "signing-api",
-  "branch": "quicksign",
+  "branch": "jam/quicksign-draft-2",
```
mightybyte: Can move this to master now.

mightybyte (approved):

PR #757, opened 2022-03-29
Token support

ohri-anurag:
This PR adds the following:
1. Tokens are now persistent.
2. They are also shown in the Switch Token dialog, similar to the UI in the Networks Dialog.
3. Users can also delete tokens (except for the KDA coin token). They can also add new tokens, although they are not validated right now, and are added as is. They can be validated, since this PR uses the monadic input widget. I was unsure of the validation method to use, which is why I took this way out, where the validation always passes.


jmininger:
Closing in favor of #767

PR #765, opened 2022-04-18
gtransfer: dont sign every metadata change

jmininger:
Currently GTransfer has two tabs: the Metadata and the Signing tab. When the user changes values in the metadata tab, they change the transfer's payload, which is represented as a Dynamic. We currently produce a new signature every time this dynamic changes. Signatures are an expensive operation (especially right now, in the browser), and this can result in frozen / glitched screens when the user attempts to change a value such as the gas-payer account. '

We only need to generate a transfer's signature after the payload is confirmed in the metadata tab and the user switches over to the signing / external signature tab. This PR addresses this.

### DIFF
--- a/frontend/src/Frontend/UI/Transfer.hs
+++ b/frontend/src/Frontend/UI/Transfer.hs
@@ -720,12 +720,20 @@ transferDialog model netInfo ti ty fks tks unused = Workflow $ do
       (conf, meta, destChainInfo) <- tabPane mempty currentTab TransferTab_Metadata $
         transferMetadata model netInfo fks tks ti ty
       let payload = buildUnsignedCmd netInfo ti ty <$> meta
-      edSigned <- tabPane mempty currentTab TransferTab_Signatures $ do
-        networkView $ transferSigs
-          <$> (model ^. wallet_keys)
-          <*> payload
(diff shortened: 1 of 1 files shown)


frontend/src/Frontend/UI/Transfer.hs
```diff
@@ -720,12 +720,20 @@ transferDialog model netInfo ti ty fks tks unused = Workflow $ do
       (conf, meta, destChainInfo) <- tabPane mempty currentTab TransferTab_Metadata $
         transferMetadata model netInfo fks tks ti ty
       let payload = buildUnsignedCmd netInfo ti ty <$> meta
-      edSigned <- tabPane mempty currentTab TransferTab_Signatures $ do
-        networkView $ transferSigs
-          <$> (model ^. wallet_keys)
-          <*> payload
-          <*> (_transferMeta_sourceChainSigners <$> meta)
-      sc <- holdUniqDyn . join =<< holdDyn (constDyn Nothing) (Just <$$> edSigned)
+      let isSignatureTab = (== TransferTab_Signatures) <$> updated currentTab
```
mightybyte, line 723: It looks like this function is unused?

mightybyte:
This was included in #766

PR #783, opened 2022-09-14
Update quicksign

jmininger:
Bump quicksign to support serialization of `SignatureList` and add CommandSigData parsing support to SigBuilder


frontend/src/Frontend/UI/Dialogs/Signing.hs
```diff
@@ -154,30 +154,45 @@ uiQuickSign
...
+uiQuickSign ideL (qsr, writeSigningResponse) onCloseExternal = (mempty, ) <$> do
+  let sendResp = performEvent . fmap (liftJSM . writeSigningResponse . Right) -- We use Either for legacy reasons, but the quicksign protocol now allows for error handling in its response, so we just `const Right` now
+      emptyReqResponse = do
+        eClose <- failureModal "QuickSign request was empty" "" onCloseExternal
+        sendResp $ QSR_Error QuicksignError_EmptyList <$ eClose
+  if _quickSignRequest_csds qsr == []
+    then emptyReqResponse
+    else case partitionEithers $ fmap snd sdReqs of
+       ([], reqs) -> quickSignModal ideL sendResp reqs onCloseExternal
+       otherwise -> do
```
mightybyte: Does this type check?  Seems like it shouldn't given https://hackage.haskell.org/package/base-4.17.0.0/docs/Prelude.html#v:otherwise

jmininger: Yes, though i can change it to `_`

mightybyte (approved):

PR #795, opened 2022-12-21
Change needed to build on macOS 10.15

jwiegley:
Note that I'm well aware I am editing a file that asks not to be modified, but Obelisk gives us no other way to vary the resolution of the thunk, since it itself makes an explicit path reference to `.obelisk/impl`.


sirlensalot (approved):

jwiegley:
I will change this PR so that the `if` test is inside the obelisk branch that I'm using, rather than happening the way it does here.

